PR #3, opened 2014-07-04
Added server side paging 

joellanciaux:
I've put together an initial version of server-side grabbing of data for Griddle so that all data doesn't have to be loaded at the same time.  This resulted in a bit of refactoring, so please review.


edaloicaro18:
Is there any example for this feature? I can't find anything in the documentation.

PR #59, opened 2015-01-23
Added sortable property to column metadata

rtsao:
It would be really helpful to have only certain columns be sortable. I'm thinking this should be a property in columnMetadata items:
```
[
  {
      "columnName": "name",
      "sortable": false,
      "displayName": "Name"
  },
  {
      "columnName": "location",
      "sortable": true,
      "displayName": "Location"
  },
  ...
]
```

### DIFF
--- a/build/griddle.js
+++ b/build/griddle.js
@@ -1118,17 +1118,6 @@ return /******/ (function(modules) { // webpackBootstrap
 	            var sortComponent = null;
 	            var titleStyles = null;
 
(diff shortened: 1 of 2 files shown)


rtsao:
Issue: https://github.com/DynamicTyped/Griddle/issues/60

ryanlanciaux:
Thanks so much for adding this!! :+1:

tony-kerz:
awesome, should be added to documentation :ghost:

PR #61, opened 2015-01-26
Lenient peer dependency on underscore.

benmosher:
`~1.6.0` rejects underscore v1.7.0. Peer dependencies [need to be lenient](http://blog.nodejs.org/2013/02/07/peer-dependencies/), otherwise this needs to be a f'real dependency, as I understand it.

Updated to `1.x` to allow future versions <2.0.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "pagination",
     "sort"
   ],
-  "main": "modules/griddle.jsx.js",
+  "main": "build/griddle.js",
   "peerDependencies": {
     "react": "~0.12.x",
     "underscore": ">= 1.6.0 < 2"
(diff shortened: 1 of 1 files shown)


benmosher:
Upon reflection, changed to `>= 1.6.0 < 2` to ensure users have at _least_ v1.6.x, as was previously stated.

joellanciaux:
Thanks a ton for putting this together. :+1: 

I reverted the 'main' update that you made due to some issues that we were experiencing in gulp/grunt builds, but we'll be looking to put that in once we have a little more testing around it.

benmosher:
No problem. Sorry about that last commit--didn't mean for it to make it into the PR, just some goofing I was doing to try to get Browserify to take. Should've branched first, heh.

PR #71, opened 2015-02-05
Use '...this.props' to simplify passing down props

danawoodman:
React has a `{...this.props}` shortcut to pass down props to other components. I've updated the code to remove some of the unnecessary duplication and simplified some of the prop formatting to be more readable. 

Happy to discuss the pros/cons and update as needed.

### DIFF
--- a/scripts/griddle.jsx
+++ b/scripts/griddle.jsx
@@ -429,17 +429,29 @@ var Griddle = React.createClass({
         };
     },
     getFilter: function(){
-     return ((this.props.showFilter && this.props.useCustomGridComponent === false) ? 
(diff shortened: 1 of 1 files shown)


ryanlanciaux:
Hey @danawoodman -- this is a great idea and something I would normally be all about accepting. We are currently going a slightly different route with passing props around. Instead of passing around a million props from the parent component to all the sub-components, we're going to pass them around in different objects. What I mean by this is that Griddle will wrap like props in their own object and pass those around. 

Please take a look at [this object](https://github.com/DynamicTyped/Griddle/blob/sixtofive/scripts/columnProperties.js) as an example of what we're currently exploring. This isn't 100% the final direction but kind of a start. Please let me know if you have any questions or thoughts on this :+1:

PR #72, opened 2015-02-05
Add missing packages for test suite to run

danawoodman:
There were a bunch of missing packages that are require to run `jest` which I've added in.

I've updated the `npm test` script to use a relative reference to `jest` so that we can fix the jest version and don't require global package installation.

I've also added an `.editorconfig` so that editors that support it will properly format the code.


danawoodman:
@ryanlanciaux do you think this is mergable? I assume the packages were missing on accident

danawoodman:
As far as I can tell, all the missing packages have been added and now `grunt` works on a fresh checkout. What are your thoughts @ryanlanciaux on also adding in an `npm run watch` task or something similar to fire up grunt? Let me know if you see things missing on the PR

PR #73, opened 2015-02-05
Add headerClassName and headerStyles props to Griddle and GriddleTitle.

### LINKED ISSUE #70: Grid table header class not working
It appears that the reference to `this.titleStyles` in the `gridTitle.jsx` file is not defined:
https://github.com/DynamicTyped/Griddle/blob/abd9319aa1bbab8c3edca0073dbb00c4a7e18858/scripts/gridTitle.jsx#L82

I'm wondering if this was intended to be a prop or possibly a computed value that just was never set. Right now it's just set to `class="undefined"` once rendered.

My proposal would be to create a `tableHeaderClassName` prop that can be passed into `<Griddle>` so that we are able to override the default table header row class. Perhaps also a `tableHeaderStyles` as well for those that like inline styles.

danawoodman:
This should fix #70 as soon as #71 is merged with some minor modifications. I've added a `headerClassName` and `headerStyles` property to Griddle so that someone can override the default header styling. 

Happy to hear feedback on implementation but it's pretty straight forward.

### DIFF
--- a/scripts/gridTitle.jsx
+++ b/scripts/gridTitle.jsx
@@ -17,7 +17,9 @@ var GridTitle = React.createClass({
            "sortDescendingClassName": "sort-descending",
            "sortAscendingComponent": " ▲",
            "sortDescendingComponent": " ▼",
-           "enableSort": true,  
+           "enableSort": true,
+           "headerClassName": "",
+           "headerStyles": "",
            "changeSort": null
         }
     },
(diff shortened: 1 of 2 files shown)


ryanlanciaux:
Thanks! :+1:

danawoodman:
Thank YOU!

danawoodman:
@ryanlanciaux does this warrant a point release? 

Also, would there be any interest in adding me as a collaborator to the project? Happy to discuss, you can email me at dana at danawoodman dot com.

PR #76, opened 2015-02-08
Allow "className" to be set on the filter component

### LINKED ISSUE #67: Remove 'row' class from filter-container when using useGriddleStyles=false
The 'row' class that is added to the GridFilter's rendered element is unfortunate IMHO. In my case, I use Bootstrap as CSS framework, and the 'row' class is part of Bootstrap's grid system. Having this class on the filter-container will therefore conflict with my grid-setup.

This class ought to not be there when using useGriddleStyles=false, in my opinion.

I have dealt with this temporarily by manually removing the class upon component mount:
```javascript
    componentDidMount: function () {
      $(this.getDOMNode()).find('.filter-container').removeClass('row');
    }
```

danawoodman:
I've added in the ability for a user to set the `className` for the wrapper HTML which should resolve #67.

I've added two test suites that should test both the default `className` and overriding it. 

*Note: I'm not sure if the default classes should be set where they are (right now in the `Griddle` and `GriddleFilter` components) or if we should set anything at all, please advise*

Let me know if I've missed something or if there is anything that should change.

### DIFF
--- a/build/griddle.js
+++ b/build/griddle.js
@@ -79,6 +79,10 @@ return /******/ (function(modules) { // webpackBootstrap
 	var _ = __webpack_require__(3);
 
 	var Griddle = React.createClass({displayName: "Griddle",
+	    propTypes: {
+	        filterClassName: React.PropTypes.string,
+	    },
+
 	    getDefaultProps: function() {
 	        return{
 	            "columns": [],
@@ -92,6 +96,7 @@ return /******/ (function(modules) { // webpackBootstrap
 	            "customRowComponentClassName":"",
 	            "settingsText": "Settings",
 	            "filterPlaceholderText": "Filter Results",
+	            "filterClassName": "row filter-container",
 	            "nextText": "Next",
 	            "previousText": "Previous",
 	            "maxRowsText": "Rows per page",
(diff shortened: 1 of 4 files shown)


build/griddle.js
```diff
@@ -79,6 +79,10 @@ return /******/ (function(modules) { // webpackBootstrap
 	var _ = __webpack_require__(3);
 
 	var Griddle = React.createClass({displayName: "Griddle",
+	    propTypes: {
```
danawoodman: @ryanlanciaux note that I've added in `propTypes` but only for the new property. Would be nice to have these set for all properties in the future. I'm happy to add in if you'd like.

build/griddle.js
```diff
@@ -494,9 +499,16 @@ return /******/ (function(modules) { // webpackBootstrap
 	        };
 	    },
 	    getFilter: function(){
-	     return ((this.props.showFilter && this.props.useCustomGridComponent === false) ? 
-	        React.createElement(GridFilter, {changeFilter: this.setFilter, placeholderText: this.props.filterPlaceholderText}) : 
-	        "");
+	        if (this.props.showFilter && this.props.useCustomGridComponent === false) {
```
danawoodman: I've made this slightly easier to read (at least for me), if you'd like I can revert to the previous formatting.

scripts/__tests__/gridFilter-test.js
```diff
@@ -6,24 +6,38 @@ var GridFilter = require('../gridFilter.jsx');
...
-		};
-
-		var input = TestUtils.findRenderedDOMComponentWithTag(filter, 'input');		
-		React.addons.TestUtils.Simulate.change(input, someEvent);
-
-		expect(mock.mock.calls).toEqual([["hi"]]);
-	})
-});
\ No newline at end of file
+    var filter;
```
danawoodman: Sorry for the confusing diff but the whitespace was updated based on the `.editorconfig`. Seems like there is a mix of tabs and spaces throughout the project.

danawoodman:
I consider this a first pass because I'm not sure the best approach for setting the defaults.  What should happen if `useGriddleStyles={false}`? Should we be setting the  `filterClassName` in the `Griddle` component or the `GriddleFilter` component (I opted for both)?

danawoodman:
Also, should I add `filterClassName` to the [properties reference page](http://dynamictyped.github.io/Griddle/properties.html) as well?

danawoodman:
@ryanlanciaux any chance we could merge this soon? Anything missing?

ryanlanciaux:
@danawoodman Sorry - I have been really swamped this week. I hope to get a better look at this this weekend.

danawoodman:
@ryanlanciaux any chance to check it out? Also, don't want to be bothersome but would you be interested in adding me as a collaborator so I can help out with the project? Thanks!

PR #90, opened 2015-02-17
Add per column filtering for local data

bparkccrl:
Often it is useful to have the ability to filter data per column via text or pre-populated lists. This change allows per column filtering to be enabled and various options (such as text or select input) to be used.


joellanciaux:
We've refactored a few things surrounding column settings/metadata, so I'm working this in _in_ a [new branch](https://github.com/DynamicTyped/Griddle/tree/columnSorting). Column filtering should be making its way into master pretty soon.

Thanks for putting this together - this is ridiculously awesome! :+1:

PR #97, opened 2015-02-20
Use type=search for better UX when filtering

tamagokun:
Using `type="search"` in some browsers will give the input an "x" button to clear the search results.


kevinold:
Thanks for your PR and support of Griddle, but we do not feel this is patch should be merged.

The main reason is that the detection, triggering and response of the "search" event type is relatively new and likely not supported consistently across browsers.

Further explanation and discussion here for reference: http://stackoverflow.com/questions/2977023/how-do-you-detect-the-clearing-of-a-search-html5-input

PR #99, opened 2015-02-23
Column sorting

bparkccrl:
* Include columnSetting.js file in the grunt copy file process.
* Modify ColumnMetadataValues so they will be updated when props are changed (such as loading results after mount).


kevinold:
@bparkccrl Thanks for this PR.  Please provide tests before this can be approved and merged.  Thanks.

kevinold:
@bparkccrl Also, please provide some background on your use case and reasoning for this patch.

bnchdrff:
@bparkccrl Thanks for working on this; I'm tackling the same issue! Did you edit `build/Griddle.js` directly? If not, do you have diffs for your changes to the source files?

PR #102, opened 2015-02-27
Add support for empty columns so items will be rendered under the matching column. 

saary:
Currently missing fields cause the grid to render values in offset under the wrong columns.
This change introduces default empty values for missing columns.

A new slim data view was added which holds default values for any missing columns and does not alter the original data.

### DIFF
--- a/build/griddle.js
+++ b/build/griddle.js
@@ -90,7 +90,7 @@ return /******/ (function(modules) { // webpackBootstrap
 	        return {
 	            columns: [],
 	            columnMetadata: [],
-	            rowMetadata: [],
+	            rowMetadata: null,
 	            resultsPerPage: 5,
 	            results: [], // Used if all results are already loaded.
 	            initialSort: "",
@@ -828,7 +828,7 @@ return /******/ (function(modules) { // webpackBootstrap
 
 	var RowProperties = (function () {
 	  function RowProperties() {
-	    var rowMetadata = arguments[0] === undefined ? [] : arguments[0];
+	    var rowMetadata = arguments[0] === undefined ? {} : arguments[0];
 	    _classCallCheck(this, RowProperties);
 
 	    this.rowMetadata = rowMetadata;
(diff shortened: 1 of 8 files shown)


saary:
Test actually passed when I ran locally. No idea why it failed on Travis.

ryanlanciaux:
Thanks! I will try locally too

ryanlanciaux:
Interesting - Travis can't find Jest for some reason on the PR. 

Strange that [Build 53](https://travis-ci.org/DynamicTyped/Griddle/builds/52563682) worked and [Build 54](https://travis-ci.org/DynamicTyped/Griddle/builds/52567183) didn't.

saary:
Indeed. Is there a way to retry a Travis build?

ryanlanciaux:
Thanks! Nice that Travis started working again too :) :+1:

saary:
:+1: Thanks!

sanketmeghani:
@saary @ryanlanciaux Is there any change in above behavior? I am using griddle-react@0.6.1 and react@15.4.1. It does not show empty column if corresponding key does not exists in the results.

`        var fakeData =  [{
            "id": 0,
            "name": "Mayer Leonard",
            "city": "Kapowsin",
            "state": "Hawaii",
            "country": "United Kingdom",
            "company": "Ovolo",
            "favoriteNumber": 7
        }, {
            "id": 1,
            "name": "Mayer Leonard",
            "city": "Kapowsin",
            "state": "Hawaii",
            "favoriteNumber": 7
        }];`

and 
`<Griddle results={fakeData}/>`

Shows value '7' in 5th column of second row instead of 7th column.

saary:
@sanketmeghani as far as I remember you need to provide the column names and this fixed made sure that columns names that did not have matching keys appeared empty. In your example you are not providing the column names.

sanketmeghani:
@saary Thank you for your response. It seems it is an issue with version 0.6.1. It did not work for me even after providing column names with griddle-react@0.6.1. It works after upgrading griddle-react to latest version @0.7.1.

saary:
Good to know. Thank you.

Issue #104, opened 2015-03-02
[Performance] All rows are re-rendered in the actual DOM whenever the application state changes


ohana54:
Hi,

I discovered a major performance issue when I tried to use the grid in my application with about 1000 rows (max results set to 10, infinite scrolling enabled).
The rows are being re-rendered (in the actual DOM) whenever any state in the application changes.
This is due to a new unique key that's being generated every time a row is rendered, so react thinks it's a new row every time and replaces everything in the actual DOM.

A reproduction is shown here: http://jsfiddle.net/u0ue24ea/
Open the console and click on the "inc" button. I print to the console the DOM changes and you can see that all the rows are being replaced.

I believe this should be addressed ASAP if people are going to rely on this grid for their apps.

Thanks!

FoxxMD:
I think this also prevents further customization of #75. When I render a custom row and then change a property of its state the row is immediately re-rendered with initial data, preventing any kind of internal interaction with the custom components.

joellanciaux:
@ohana54 Thanks for putting together an issue. We added only rendering relevant rows infinite scrolling in master a little bit ago, but it hasn't made it's way to npm just yet. We're hoping to get this out with some other features/improvements in the near future, but you might be able to get moving with the source - I'll write up a little more on this later tonight. 

This initially came about because we were a little reluctant to specify a row height when infinite scrolling, but I think that's the only way to make this happen.

scvnc:
I have a hunch that the issue is [usage of `_.uniqueId`]
(https://github.com/DynamicTyped/Griddle/search?utf8=%E2%9C%93&q=uniqueId&type=Code) for keying of components internally. I had similar problems with a different react grid doing the same thing. I was able to fix this when I removed usage of `_.uniqueId` in favor of requiring the user to define which property on the row object can uniquely identify a row. Otherwise every row was looking like a entirely different row since the keys get incremented every time react calls `render()`

After that it was able to properly diff and manipulate the DOM more efficiently.  Using `_.uniqueId` may be a React anti-pattern.

ohana54:
@joellanciaux I may have not been clear, the issue is present regardless of infinite scrolling. My example shows a basic grid without any fancy options. All the rows are re-rendered when any *not grid related* state changes. Again, I show it in the example.

@scvnc Yes, that is the root cause. I fixed it as well in the same way you did, but still experienced slow rendering (~250ms), which sent me looking for alternatives.

scvnc:
[...]

PR #105, opened 2015-03-02
Updated grid to allow user to send in unique ID field via rowMetadata

blkmutt:
This should provide consistent keys based on a column in the table to help with #104.

### DIFF
--- a/build/griddle.js
+++ b/build/griddle.js
@@ -71,23 +71,26 @@ return /******/ (function(modules) { // webpackBootstrap
 	   See License / Disclaimer https://raw.githubusercontent.com/DynamicTyped/Griddle/master/LICENSE
 	*/
 	var React = __webpack_require__(2);
-	var GridTable = __webpack_require__(5);
-	var GridFilter = __webpack_require__(6);
-	var GridPagination = __webpack_require__(7);
-	var GridSettings = __webpack_require__(8);
-	var GridNoData = __webpack_require__(9);
-	var CustomRowComponentContainer = __webpack_require__(10);
-	var CustomPaginationContainer = __webpack_require__(11);
+	var GridTable = __webpack_require__(6);
+	var GridFilter = __webpack_require__(7);
+	var GridPagination = __webpack_require__(8);
+	var GridSettings = __webpack_require__(9);
+	var GridNoData = __webpack_require__(10);
+	var CustomRowComponentContainer = __webpack_require__(11);
+	var CustomPaginationContainer = __webpack_require__(12);
 	var ColumnProperties = __webpack_require__(4);
+	var RowProperties = __webpack_require__(5);
 	var _ = __webpack_require__(3);
 
 	var Griddle = React.createClass({
 	    displayName: "Griddle",
 	    columnSettings: null,
+	    rowSettings: null,
 	    getDefaultProps: function () {
 	        return {
 	            columns: [],
 	            columnMetadata: [],
+	            rowMetadata: [],
 	            resultsPerPage: 5,
 	            results: [], // Used if all results are already loaded.
 	            initialSort: "",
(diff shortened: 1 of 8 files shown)


blkmutt:
Changing the format of the rowMetadata.

scvnc:
Maybe there is some way to DRY this code up so that the config-check logic isn't repeated between gridTable.js, griddle.js, among other places row _.uniqueId was previously being used.

Also, I think that this should produce a warning that _.uniqueId is being used if no key is defined in the rowMetadata config.

(edit: posting this here instead)

joellanciaux:
Hrm, potentially moving this into `rowProperties.js` might make sense. 

Thanks a ton for putting this together @blkmutt :+1:

blkmutt:
This is just doing the check one place, griddle.js is the built version. I think it would make more sense to put the config check in gridTable.js as we're checking to make sure the Row Properties exist. Mind you, I could put the check for the properties in gridTable.js and the check for the metadata in rowProperties.js ... thoughts?

scvnc:
Ah, yeah :+1: having the build checked in to the repo is disorienting to me.

The two places I see for picking off `_.uniqueId` are gridTable.jsx and [gridRowContainer.jsx](https://github.com/BlackMuttMedia/Griddle/blob/master/scripts/gridRowContainer.jsx#L76). I assume it would be good to fix both of those.

blkmutt:
Yeah, I believe we did pass it along. I did update rowProperties.js to include the check for the key, though, so it's at least clear in code. Also, let me know if you want the phrasing of the warning to be changed.

joellanciaux:
It's already been said, but, yeah, it does need to happen on 'gridRowContainer' to when putting together rendering for subgrid rows.

Aside from it being in the two files, I think I'm a _little_ bit of an advocate for the check living in 'rowProperties', as it might allow the logic to be passed to customGrid/Row components.

scvnc:
maybe like a `var key = that.props.rowSettings.getRowKey(row)` ? Return some configured key value or the result of `_.uniqueId`.  What do y'all think?

Also note the failing travis tests that may need to be updated.

blkmutt:
I could update and put the row key logic right in there. Sounds good. Just as an FYI, if I do that, we'll have the warning for each row. Thoughts?

scvnc:
As for warning for each row, in the grander scheme of things, I use a `WarnService` piece that wraps `console.warn` and can reduce multiple `warn()` calls to a single warn. It also handles silencing of warnings when webpack builds minified production code.

This is a package maintainer decision, but if it was up to me, I'd omit the warns for now and make a github issue to handle this more robustly later.  The docs have to be updated too.

blkmutt:
Alright, logic should be in rowProperties.js and documentation should be updated.

joellanciaux:
:+1: Sweet deal.  Thanks again @blkmutt!

@scvnc Regarding wrapping up warnings in a service, that would be slick, but it might negate a little bit of the usefulness of the 'no row key' error in general use. It's definitely worth discussing, though.

ohana54:
@blkmutt Thanks for the fix! One question though, why internally are you using an array as a default value when the api expects an object? Did I miss something?

blkmutt:
No, you didn't ... I did. I'd initially had it set up like the Column Properties, but realized it just didn't make sense. I'll try to make some updates and get a fix out today.

joellanciaux:
@blkmutt @ohana54 I believe that I updated that while I was pulling it in to `master`. Did I possibly miss something? https://github.com/DynamicTyped/Griddle/blob/master/scripts/rowProperties.js

blkmutt:
Yeah, it actually looks like you did, @joellanciaux.

weitenghuang:
do you have an example how to assign unique key when using `rowMetadata` props?

weitenghuang:
When I try to have isomorphic griddle-react, i realize i will need to provide `rowMetadata={{"key":columnName}}`, otherwise _.uniqueId() seems to cause the react invalid checkSum issue between server and client rendering. Maybe we can add this to griddle's document?

PR #119, opened 2015-03-25
sort the columns in grid settings

vith:
so they appear in the same order there as in the table


ryanlanciaux:
Thanks a ton! :+1:

Issue #64, opened 2015-01-31
Row events


wavded:
Apologies if I missed this in the documentation but I'd like a way to have events when rows are interacted with.  Namely if they are clicked or hovered over.  Is this doable?  Thanks!

ryanlanciaux:
Thanks for posting this! Currently this isn't something that can be done easily but we are adding some ways to easily have events on these items.

hc000:
I was playing around with this and I used the useCustomRowComponent and customRowComponent to add another component (and add the event in the other components).

However it removed the default formatting (column header and spacings) of the original table.

Any suggestions on how to go about and integrate them together?

<Griddle results={fakeData} tableClassName="table" showFilter={true} useCustomRowComponent={true} customRowComponent={Popup2}/>

var Popup2 = React.createClass({
  getDefaultProps: function(){
    return { "data": {} };
  },
  showSimpleDialog: function(e){
      this.refs.simpleDialog.show();
  },
  render:function(){
    return (
      <div>
          <div onMouseDown={this.showSimpleDialog}>{this.props.data}</div>
        <SkyLight ref="simpleDialog" >
          I'm a popup!
        </SkyLight>
      </div>
    )
  }
});

nextminds:
I've currently added an 'onRowClick' prop to Griddle, and then pass that down to gridTable, gridRowContainer and then gridRow. The callback receives the row data and a few other props. Would this be an acceptable feature for a pull request?

ryanlanciaux:
That would be excellent :+1:

bpugh:
Just checking on the status of this. @nextminds your changes look good to me. Any reason you haven't submitted a PR?

ryanlanciaux:
All, we're closing this in favor of https://github.com/GriddleGriddle/griddle-render/issues/15 that we're treating as acceptance criteria for the more modularized components that we're working towards. Please feel free to let us know of any questions :+1:

doubledherin:
I'm confused -- the Properties documentation shows onRowClick, but it seems like it's not actually implemented yet. What's the scoop?

PR #122, opened 2015-04-03
Add Row Events: onRowClick #64

nextminds:
Added onRowClick event to Griddle, passing it down to GridRow. The callback receives the gridRow component and the event. The gridRow is and instance of GridRow which has a 'props' property which has 'data', 'columnSettings', 'rowSettings', etc. This does not consider child-tables, yet. Usage example:

``` javascript
function(gridRow, event) {
    console.log("Grid row clicked with data:", gridRow.props.data);
}
// ...
<Griddle onRowClick={onRowClick} />
```


joellanciaux:
Thanks for putting this together! :+1:

PR #127, opened 2015-04-16
Don't rely on global Griddle

vkurchatkin:
Not sure if this is the best solution, but it work and I don't want to mess with circular dependencies


ryanlanciaux:
Thanks!

PR #128, opened 2015-04-16
Avoid passing userGriddleStyles as style prop

vkurchatkin:
Apparently passing `false` as a value of `style` prop is not working


joellanciaux:
Good catch and thanks for the fix! :+1:

PR #134, opened 2015-04-26
Add custom table row component

ryanlanciaux:
This allows custom table rows to be created. Documentation still needs to be updated and it would likely be good to state that the names on the custom components will likely change before ver 1 is out (aka customRow and customTableRow components are kind of confusing names).

A rough example of a `customTableRowComponent` is as follows: 

``` 
   var CustomRow = React.createClass({
      getInitialState: function() {
        return { editToggled: false };
      },
      handleClick: function(e){
        this.setState({editToggled: !this.state.editToggled });
      },
      render: function(){
        if(this.state.editToggled){
          var cols = _.map(_.pairs(this.props.data), function(col, index){
            return <td><input type="text" defaultValue={col[1]} style={{width: "100%", height: "100%", color: "#CCC"}} />{index === _.pairs(this.props.data).length -1 ? 
              <button onClick={this.handleClick} style={{position: "absolute", right: 0}}>Done</button> : null}</td>
          }, this);
        }
        return this.state.editToggled ?
          <tr>{cols}</tr> :
          <Griddle.GridRow {...this.props} onRowClick={this.handleClick} />;
      }
    });
```

### DIFF
--- a/docs/assets/index.html
+++ b/docs/assets/index.html
@@ -254,7 +254,7 @@ var Griddle = require('griddle-react');</pre></code>
     var CustomRow = React.createClass({
       render: function(){
         var cols = _.pairs(this.props.data).map(function(col, index){
-          return <td>{col[0]}</td>
+          return <td><input type="text" defaultValue={col[1]} style={{width: "100%", height: "100%", color: "#CCC"}} /></td>
         });
         return (
           <tr>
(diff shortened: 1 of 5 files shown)


6ewis:
What's the status?

israelidanny:
@ryanlanciaux Is it just me or does this merge use either lodash or underscore without actually require-ing it?

Jhawins:
This never happened either.

emorling:
Does this allow easy editing of the underlying data?

PR #137, opened 2015-04-27
Switch to TFOOT for pagingContent issue #177

mattcodez:


kevinold:
@mattcodez Please describe the benefits and reasoning behind switching this tag.  Need justification for this and additional background before accepting.  Thanks!

yordis:
:+1: @kevinold semantically is correct and even from styling standpoint is better #182

Issue #120, opened 2015-03-28
require("_") in build/griddle.js isn't compatible with lodash


Firfi:
There could be situation when griddle.js code that run on window object have lodash' `_` instead underscore `_`. Then, when you call `.rest` on this object, results are different. Solution can be calling _.drop instead of _.rest like this 

```
_.rest([1,2], 0)
[2]
_.drop([1,2], 0)
[1, 2]
```

or use `underscore` in build explicitly.

bizmurr:
:+1:

https://github.com/GriddleGriddle/Griddle/blob/06f4ea942af0abe0203a1e62089e4756efa4e0e1/scripts/griddle.jsx#L352

https://github.com/GriddleGriddle/Griddle/blob/06f4ea942af0abe0203a1e62089e4756efa4e0e1/build/griddle.js#L413

Are these the only places where it needs to be changed?

Firfi:
It doesn't seems so. I changed only these two lines in my 'branch' [not branch really, there's reasons] and use Griddle without problems.

ryanlanciaux:
Sorry for the delay in response to the initial issue. I agree that it would be good to allow for underscore or lodash -- we are likely switching to lodash (hopefully give those using an underscore a way to continue using Griddle+Underscore) in v0.3.0 / 0.4.0 :+1:

Firfi:
> underscore backward compat

Looks like that's more than possible, however, conditional `||` needs to be included in code as these functions have different meaning for lodash and underscore. See my linked commits.

PR #143, opened 2015-04-30
Switched from _.rest to _.drop

Tarun17:
This should fix https://github.com/GriddleGriddle/Griddle/issues/120


ryanlanciaux:
Thanks for adding this! :+1:

PR #145, opened 2015-05-01
Multiple selection

cosminnicula:
- does not support subgrids
- the selection is reset when sorting / paging / filtering
- the selection is not reset when paging with infinite scroll

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ docs/html
 
 # Users Environment Variables
 .lock-wscript
+
+.idea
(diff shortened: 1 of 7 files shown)


bjrmatos:
cool!!!

ryanlanciaux:
Thanks so much! This is quite nice! :+1:

bjrmatos:
@ryanlanciaux i've just found that React throws a warning when using this new feature.. and that is because the new column (checkbox column) not have a key property.

bjrmatos:
warnings

bjrmatos:
Also this breaks support for React 0.12.x, because of the use of `React.findDOMNode` in `handleSelectClick` method in `gridRow.jsx file

ryanlanciaux:
Thanks for the heads-up @bjrmatos -- that _should_ be fixed now :+1:

PR #166, opened 2015-05-25
Add Griddle to bower

MMore:
I added the `bower.json` file to just add it to bower. You have to register it afterwards as bower component. This allows people to integrate Griddle much easier when not using NPM.

For instance I will use it with [Rails Assets](https://rails-assets.org).


ryanlanciaux:
Thanks!

PR #176, opened 2015-06-24
Add row enter/leave animation support via ReactCSSTransitionGroup 

saary:
1. Added dependency on [rc-css-transition-group](https://github.com/react-component/css-transition-group) for proper React animation support.
2. Updated the docs accordingly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,6 @@ build/Release
 # Commenting this out is preferred by some people, see
 # https://npmjs.org/doc/faq.html#Should-I-check-my-node_modules-folder-into-git
 node_modules
-modules
 docs/html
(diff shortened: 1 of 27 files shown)


kevinold:
@saary Please add tests before this can be officially approved and merged.  Thanks.

rashthedude:
What happened to this merge, any updates?

html-monster:
https://github.com/html-monster/ReactUpdateAnimation
https://html-monster.github.io/ – Demo
I’ve modified an extension of ReactCSSTransitionGroup component. It expands component capabilities, now the class will be added not only while the appearance or removal of an element, but also when an item is updated

PR #188, opened 2015-07-23
Document how to style an individual row.

arianitu:
This is an undocumented feature that allows you to style a row based on the rows data.

### DIFF
--- a/docs/src/markdown/customization.md
+++ b/docs/src/markdown/customization.md
@@ -200,22 +200,24 @@ React.render(<Griddle data={someData} columnMetadata={columnMeta} />,
 #####Example:#####
 
 ```javascript
-	  var rowMetaData = {
-		"bodyCssClassName": function(rowData) {
-		  if (rowData.action === "added") {
-			    return "green-row";
-		  } else if (rowData.action === "removed") {
-			    return "red-row";
-		  } else if (rowData.action === "transfer") {
-			    return "blue-row";
-		  }
-	  };
+var rowMetaData = {
+    "bodyCssClassName": function(rowData) {
+        if (rowData.action === "added") {
+            return "green-row";
+        } else if (rowData.action === "removed") {
+            return "red-row";
+        } else if (rowData.action === "transfer") {
+            return "blue-row";
+        }
+        return "default-row";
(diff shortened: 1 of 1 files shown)


ryanlanciaux:
Thanks so much!

yordis:
@arianitu yeah man, thanks a lot! gg

hedgerh:
@arianitu i followed these instructions, but nothing seems to happen.  my function is never called, and the rows default to the 'standard-row' class.  im going to dig into the code to see whats up if you aren't around atm.  here's my render method:

```
  render() {
    const rowMetaData = {
      'bodyCssClassName': function(rowData) {
        return 'pizza-row';
      }
    };

    return (
      <div className='AudioList'>
        <Griddle
          results={ this.props.tracks }
          resultsPerPage={ this.props.tracks.length }
          columns={ this.props.columns }
          showPager={ false }
          onRowClick={ this.props.onRowClick }
          rowMetaData={ rowMetaData } />
      </div>
    );
  };
}
```

Griddle version:
```
"griddle-react": "^0.2.13",
```

Thanks for all your work!

cc @ryanlanciaux @yordis

hedgerh:
@arianitu: An update on my issue, I didn't realize that the prop was "rowMetadata" with a lowercase 'd'.  It was actually really easy to overlook too.  Maybe that should be addressed?

yordis:
It shouldn't

hedgerh:
@yordis: I'm sorry, I'm not sure what you mean.  My issue was that I was passing Griddle:

```
rowMetaData={ rowMetaData }
```

instead of:

```
rowMetadata={ rowMetaData }
```

I just wanted to mention that I had a pretty hard time noticing what was wrong.  The solutions would be to either rename the prop "rowMetaData" or change the example to use var rowMetadata = { ....

Just a suggestion!  If either of those sound good, I can submit a PR.

yordis:
@hedgerh I think Metadata is the correct way to put it, unless I am incorrect about English rule.

hedgerh:
@yordis You are correct.  https://github.com/GriddleGriddle/Griddle/pull/195

Issue #198, opened 2015-08-12
Custom components should have more control over the props


yordis:
[Because we don't pass anything else than the results and className](https://github.com/GriddleGriddle/Griddle/blob/61807c8ecfb07a497618174b98cafb3beec30f71/scripts/griddle.jsx#L652) you don't have too much control over the props and in my case I want to use something like ```onSelect``` event so I need to add some function in the Griddle component and use it inside the custom grid component.

Probably we should add more props or even add all the props to the custom grid component.

PS: You can't use onRowClick click either.

yordis:
@ryanlanciaux @kevinold I really need this, I can do a PR if you want

ryanlanciaux:
Howdy @yordis. We totally agree that there needs to be more control over the custom components. I'm on board with a PR, however there is the possibility that we may be working on duplicate functionality between this and the griddle-render project. Have you had a chance to look at it? It takes in all the components as props to griddle and every part can be overridden (with the ability to specify any event, etc).

Please let me know and thanks for the great suggestion :+1:

yordis:
@ryanlanciaux very smart movement in separated the component in two modules :+1:  I got what you did there. So let ask you something. How can I build the project using griddle-core and griddler-render? 

Right now I sort of understand what you did but I don't have any idea about how to use it, if you can provide any route for use the other concept will be great

@ryanlanciaux which one is the best method for communicate with the your team? I don't want to create issues for some discussion about griddle.

eriknyk:
+1

McChen:
+1

PR #200, opened 2015-08-13
Added gridMetadata

yordis:
@ryanlanciaux  I understood discussion about duplicate functionality but the griddle-react package in npm needs this, like me right now. 

This is a very simple implementation, you pass all the data throw ```gridMetadata``` to the custom grid component. The only concern I have is about passing ```data``` or ```className``` because I didn't pass the metadata throw some scope like ```gridData``` just because I didn't like to create a extra scope for avoid these mistake, however, this is easy to change and depend of you.

About unit testing, you are using ```jest```, jest using ```iojs``` instead of node. I could't get it run properly. 
My suggestion is try to choose stack that people know, the big problem with JS is that one, too many packages with the same objective and I can't deal with everything, even was ```iojs``` problems which I have no idea what is the different from nodejs or even if I have to create a new environment for him because I can get node conflicts. I hope you understand my point

Related to https://github.com/GriddleGriddle/Griddle/issues/198

### DIFF
--- a/scripts/griddle.jsx
+++ b/scripts/griddle.jsx
@@ -32,6 +32,7 @@ var Griddle = React.createClass({
     getDefaultProps: function() {
         return{
             "columns": [],
+            "gridMetadata": null,
             "columnMetadata": [],
             "rowMetadata": null,
             "resultsPerPage":5,
@@ -649,7 +650,7 @@ var Griddle = React.createClass({
         ) : "";
     },
     getCustomGridSection: function(){
-        return <this.props.customGridComponent data={this.props.results} className={this.props.customGridComponentClassName} />
+        return <this.props.customGridComponent data={this.props.results} className={this.props.customGridComponentClassName} {... this.props.gridMetadata} />
     },
     getCustomRowSection: function(data, cols, meta, pagingContent){
         return <div><CustomRowComponentContainer data={data} columns={cols} metadataColumns={meta}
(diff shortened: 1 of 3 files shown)


yordis:
@ryanlanciaux please let me know if this can be in in the next release. I really need this

scripts/griddle.jsx
```diff
@@ -649,7 +650,7 @@ var Griddle = React.createClass({
         ) : "";
     },
     getCustomGridSection: function(){
-        return <this.props.customGridComponent data={this.props.results} className={this.props.customGridComponentClassName} />
+        return <this.props.customGridComponent data={this.props.results} className={this.props.customGridComponentClassName} {... this.props.gridMetadata} />
```
ryanlanciaux: Should this be spread operator? I think you could get away with just using `this.props.gridMetadata`. Also it looks like it has a space before `this.props.gridMetadata`.

yordis: It's spread operator. I should remove that space there.

I don't use `this.props.gridMetadata` because then what's the point of passing other properties like `rowClick`. You have to create another property inside the `gridMetadata` that simulate the same result. When we customize the `grid` we change everything, in my opinion, we should have all the properties for the `regular grid` + whatever you need in the `gridMetadata` but that's not a big deal we can change it.

kevinold:
@yordis Please update your branch with the latest master which includes #241. This will allow your tests to run on Travis.  Thanks!

yordis:
@kevinold I messed up something with my branch :cry:

yordis:
@kevinold done!

yordis:
@kevinold Are you gonna merge it?

PR #217, opened 2015-08-31
Update new github url in package.json

KnutHelland:


ryanlanciaux:
Thanks for changing this to point to the right place :+1:

PR #219, opened 2015-09-01
Fix invariant violation by showPager

KnutHelland:
If showPager fluxes between true and false, having `var pagingContent = ""` (which is equal to `var pagingContent = <span />;`) results in invariant violation (at least in Chrome). Change it to `<tbody />` solves all problems for me.


prycejs:
Just linking this to issue #221 as it looks to be related.

ryanlanciaux:
Thanks!

Issue #225, opened 2015-09-08
can not pass properties into component set for customComponent column attribute


eriknyk:
HI,

I have columns def:

```
let columns = [{
            'columnName': 'id',
            'locked': true,
            'visible': true,
            'displayName': 'ID'
        }, {
            'columnName': 'name',
            'locked': true,
            'visible': true,
            'displayName': 'Name',
            'customComponent': UserLink
        }, {
            'columnName': 'type',
            'locked': false,
            'visible': true,
            'displayName': 'Type'
        }];
```

and UserLink:

```
class UserLink extends Component {
    static propTypes = {
        rowData: PropTypes.object,
        data: PropTypes.string,
        params: PropTypes.object
    };

    render () {
        const { companyId } = this.props.params;
        const url = '/admin/companies/' + this.props.companyId + '/users/' + this.props.rowData.id + '/details';
        return <Link to={url}>{ this.props.data }</Link>;
    }
}
```

How can I pass companyId (to be used this.props.companyId ) to the UserLink component??

I tried something like this:

```

let UserLinkComp = <UserLink companyId={companyId} />

let columns = [{
            'columnName': 'id',
            'locked': true,
            'visible': true,
            'displayName': 'ID'
        }, {
            'columnName': 'name',
            'locked': true,
            'visible': true,
            'displayName': 'Name',
            'customComponent': UserLinkComp
        }, {
            'columnName': 'type',
            'locked': false,
            'visible': true,
            'displayName': 'Type'
        }];

```

But it doesn't worked, and error was thrown.

Best Regards.

prycejs:
I passed customComponent settings in the following way:

In the following JS fiddle, there is a custom component MyCustomComponent that refers to data passed through under Griddle columnMetaData prop.  There is also an example of using other fields from rowData (in this case, rowData.name)  in rendering MyCustomComponent, and passing a function to dynamically compute something based on rowData, at render time.

https://jsfiddle.net/ws9pp3nu/3/

eriknyk:
It worked,

Thank you a lot @prycejs

ruchigoyal2005:
@prycejs @eriknyk : I do see locked:true property for columns. Does it lock the column while scrolling?

PR #227, opened 2015-09-10
Updating customization.md document to include information about custom column component metadata

rovolution:
To Whomever it May Concern,

First wanted to start off by saying thank you for building this awesome plugin!

I am building a Flux app using Alt, and needed a way to pass an actions object down to a column custom component as a prop. After further examining the source code, I noticed that the `metadata` property is being passed down to the component, and the action can be specified in the `customColumnMetadata` object.

However, the documentation did not mention any of these things. This PR updates the docs to better reflect this information. Please let me know if I need to change anything.

This is related to [this issue](https://github.com/GriddleGriddle/Griddle/issues/225)


ryanlanciaux:
Thanks so much for the writing this up!

PR #235, opened 2015-09-19
Fix row per page

icharlie:
The row per page can't be changed if the amount of the results is less than the selected number.

example: https://jsfiddle.net/9nhrdmek/


kevinold:
@icharlie Please update your branch with the latest master which includes #241. This will allow your tests to run on Travis.  Thanks!

icharlie:
Done. 
I also changed the griddle-test.js to check `state.resultsPerPage` not `props.resultsPerPage` because of this pull request.

PR #243, opened 2015-09-30
Updating gh-pages with latest docs

rovolution:
This should update the docs with the latest contents from the markdown files on the `master` branch...I just ran the `grunt build` task and copied the generated contents from `/docs/html` over.

Ideally, this would be automated via a npm post-publish script or on a CI server

### DIFF
--- a/customization.html
+++ b/customization.html
@@ -85,7 +85,7 @@
 
 <h3 id="column-metadata">Column Metadata</h3>
 <p>The column meta data property is used to specify column properties that are not part of the result data object. For instance, if you want to specify a displayName that is different than the property name in the result data, the <code>columnMetadata</code> property is where this would be defined. </p>
-<p>The properties that the columnMetadata object can contain are as follows: </p>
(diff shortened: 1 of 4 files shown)


rovolution:
@ryanlanciaux have you had a chance to check this out? thanks!

ryanlanciaux:
This looks good -- do you have any thoughts around setting this up with a CI-server? I haven't done that and would love to get that rolling for the new stuff.

rovolution:
@ryanlanciaux

You could setup the CI server such that once a PR gets merged into `master` and the version is bumped/new tags are pushed to GH, it does the following:

- runs the tests/linter
  - if the tests/lint check passes, automatically run the the documentation generation task and push the results to the `gh-pages` branch
  - if the tests/link fail, do nothing  


But step 1 is to setup a CI server on the project. I've used both CircleCI and Travis in the past, though I would go with Travis since they provide free support for open-source projects. [You can read more about the various CI services here](https://strongloop.com/strongblog/node-js-travis-circle-codeship-compare/)

PR #258, opened 2015-10-19
fixes #126 Uncaught Error: Invariant Violation

### LINKED ISSUE #126: Uncaught Error: Invariant Violation when using children and dynamically changing `result`
When I have children for some of my rows, and dynamically changing input data (result=), i.e. just filtering it for some rows, I get error 

```

Uncaught Error: Invariant Violation: processUpdates(): Unable to find child 22 of element. This probably means the DOM was unexpectedly mutated (e.g., by the browser), usually due to forgetting a <tbody> when using tables, nesting tags like <form>, <p>, or <a>, or using non-SVG elements in an <svg> parent. Try inspecting the child nodes of the element with React ID `.0.2.0.0.0

```

After which data isn't filtered and child expand link cease to work.

albertolacework:
fixes #126 by sorting nodes with children and nodes without children. Originally this did not happen for whatever reason. It was either if at least one had children they were rendered as all with children or they were rendered as all without children.

### DIFF
--- a/scripts/gridTable.jsx
+++ b/scripts/gridTable.jsx
@@ -99,12 +99,13 @@ var GridTable = React.createClass({
   getAdjustedRowHeight: function() {
     return this.props.rowHeight + this.props.paddingHeight * 2; // account for padding.
   },
-  getNodeContent: function() {
+  getNodeContent: function () {
     this.verifyProps();
     var that = this;
 
     //figure out if we need to wrap the group in one tbody or many
-    var anyHasChildren = false;
+    var nodesWithChildren = [];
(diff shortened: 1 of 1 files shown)


ryanlanciaux:
This looks great -- any way you could switch it to jsx to be consistent with how the project is currently structured? Thanks!

albertolacework:
Ah I've been busy with work and it looks like I was working on an older version. I'll make a fix now.

albertolacework:
Yeah, the build failed for reasons I expected it should if the test coverage reached that point.

PR #268, opened 2015-11-09
Update React 0.14.2 + other dependencies

sajinshrestha:
* Updates
    * `React 0.14.2`
    * Replace `grunt-6to5` with `grunt-babel`
    * Replace deprecated `grunt-react` with `grunt-babel`
    * `grunt-webpack`
    * Replace deprecated `jsx-loader` with `babel-loader`


sajinshrestha:
Updating tests. Will submit another PR in a bit.

ryanlanciaux:
Awesome! This looks like great work. 

In regard to the error, I will try to take a look tonight and see if we can't get it straightened out :)

sajinshrestha:
@ryanlanciaux Super! thank you.

ryanlanciaux:
Hmm. I'm getting some different errors locally (not related to your changes). Trying to figure out what's going on here -- I will keep you posted.

sajinshrestha:
Getting following error: 

```
FAIL  __tests__/gridFilter-test.js (0.578s)
● GridFilter › it calls change filter when clicked
  - TypeError: Can't add property changeFilter, object is not extensible
```
But I can pass the test on `gridFilter-test.js` by using
```
filter.props = {
  changeFilter: mock
};
```
instead of 
```
filter.props.changeFilter = mock;
```
But the original props will be lost. Not sure how to properly use `jest` doing some reading right now.

Let me know if I can help in any way. So close :+1:

sajinshrestha:
Grunt build works as expected. 

Main problem here are the tests. Made some more updates + fixed tests (`jest`).
Using `node v4.2.2` and `npm v2.14.7`.

So far - 15 tests failed, 58 tests passed.  

Still getting the following error because of invalid html (according to `react v0.14`). 

E.g.  is missing <table>. 
```
return
    	<thead>
            <tr>
                {node}
            </tr>
    </thead>;
```
How can we just `jest` table headers? Any ideas @ryanlanciaux?

```
Error: Invariant Violation: findComponentRoot(..., .0.0.$one): Unable to find element. This probably means the DOM was unexpectedly mutated (e.g., by the browser), usually due to forgetting a <tbody> when using tables, nesting tags like <form>, <p>, or <a>, or using non-SVG elements in an <svg> parent. Try inspecting the child nodes of the element with React ID ``.
```

sajinshrestha:
All checks have passed :+1:  @ryanlanciaux

kevinold:
@sajinshrestha @ryanlanciaux I'll test in the next day.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '0.10'
+  - '4.2.2'
```
kevinold: @sajinshrestha We will likely have users on everything from 0.10.x, or at the minimum 0.12.x that we need to support.  Thoughts on adding a range of environments to test against?

sajinshrestha: @kevinold I can change the `node version` back to `0.10` or `0.12` but jest v0.5+  will not work with node 0.12. See https://github.com/facebook/jest/issues/427#issuecomment-149773152

Right now, you can still run/build (grunt) `griddle` on `node v0.10.40` with this PR, but tests will fail.

ryanlanciaux:
Nice work @sajinshrestha! Thanks a ton! :+1: :star: :star2:

jgwconsulting:
I'll second that, thanks @sajinshrestha :)

sajinshrestha:
@ryanlanciaux @jgwconsulting Glad to help out. There are still two existing issues with the documentation - https://github.com/sajinshrestha/Griddle/issues

PR #284, opened 2015-12-07
Only use default styling for CustomRowComponentContainer if useGriddleStyles is true

dlong500:
This adds a check that prevents the default container styles being applied to CustomRowComponentContainer if useGriddleStyles is false.


ryanlanciaux:
Thanks :)

PR #294, opened 2015-12-17
Subgrid Icons are not Displayed if Metadata

cmain:
Fix for:
The component or characters that are added into the first column of a subgrid row (parentRowCollapsedComponent and parentRowExpandedComponent, default or custom) are not displayed when a metadata object (from columnMetadata) exists for this first column.

### DIFF
--- a/scripts/gridRow.jsx
+++ b/scripts/gridRow.jsx
@@ -94,9 +94,13 @@ var GridRow = React.createClass({
               columnStyles = _.extend(columnStyles, {paddingLeft:10})
             }
 
-            if (this.props.columnSettings.hasColumnMetadata() && typeof meta !== "undefined"){
-              var colData = (typeof meta.customComponent === 'undefined' || meta.customComponent === null) ? col[1] : <meta.customComponent data={col[1]} rowData={dataView} metadata={meta} />;
-              returnValue = (meta == null ? returnValue : <td onClick={this.handleClick} className={meta.cssClassName} key={index} style={columnStyles}>{colData}</td>);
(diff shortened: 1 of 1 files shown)


scripts/gridRow.jsx
```diff
@@ -94,9 +94,13 @@ var GridRow = React.createClass({
               columnStyles = _.extend(columnStyles, {paddingLeft:10})
             }
 
-            if (this.props.columnSettings.hasColumnMetadata() && typeof meta !== "undefined"){
-              var colData = (typeof meta.customComponent === 'undefined' || meta.customComponent === null) ? col[1] : <meta.customComponent data={col[1]} rowData={dataView} metadata={meta} />;
-              returnValue = (meta == null ? returnValue : <td onClick={this.handleClick} className={meta.cssClassName} key={index} style={columnStyles}>{colData}</td>);
+            if (this.props.columnSettings.hasColumnMetadata() && typeof meta !== 'undefined' && meta !== null) {
+              if (typeof meta.customComponent !== 'undefined' && meta.customComponent !== null) {
+                var customComponent = React.createElement(meta.customComponent, {data: col[1], rowData: dataView, metadata: meta});
```
ryanlanciaux: Could we please keep these in jsx instead of `React.createElement`? `<td>...` etc :)

ryanlanciaux:
Thanks a ton for submitting this -- I would love to get this merged.

scripts/gridRow.jsx
cmain: Woops, my bad. Will put up a new patch.

Issue #286, opened 2015-12-14
Passing props to customRowComponent


dlong500:
I looked through the issues and didn't see anything specifically addressing my need to pass additional props to a `customRowComponent`.  For example, I have application state that needs to be referenced in the render method of a `customRowComponent`.

A simple example would be that the site could be in one of two modes (live, maintenance).  The list of items being rendered in Griddle would render differently depending on the site mode.  Right now I can't see an easy way of getting access to that data in the `customRowComponent` render method.  Obviously I could do some hacky and inefficient stuff to add fields to each row of the results after the data is retrieved from the server, but that's not pretty at all.  Is there any way to do this now?

Is there at least a way to define some specific prop like `customRowComponentMetaData` that would get passed into each row?

dlong500:
I'll probably try to add this feature myself and submit a pull request sometime over the next week.

sanju-sfo:
Thanks that's will be very valuable, I had a similar situation and ended up using context attribute.

dlong500:
@ryanlanciaux I've added this feature to a branch on my fork of Griddle, but I have a few questions about how to best handle the pull request.

First, the overview of the feature I added is a prop called `customRowExtraData` that will get passed through the component tree and into each `customRowComponent` instance (if `useCustomRowComponent` is true).  This allows a developer to assemble an object of any data that is global to the Griddle component but needs to be accessible in the `customRowComponent` instance.

Now, I had some trouble while trying to test out the changes I made.  I saw issue #100, though I didn't have quite the same problems as FoxxMD had (granted that was nearly a year ago).  However, I did realize that there were a couple things I had to do in order to have npm install my fork of Griddle directly from Github so I could test it.  Here's what I had to do:
1) Run `grunt` to generate the `build/griddle.js` file and then commit this updated file.
2) Remove the `modules` line from `.gitignore` and then commit both the updated `.gitignore` file along with the entire contents of the `modules` folder.

After this, I could now reference my fork in my `package.json` file as below (where <commit-hash> is the hash of the specific commit at the desired point of the repo):

    "griddle-react": "git+https://github.com/dlong500/Griddle.git#<commit-hash>"

This worked for me to test things out, but now I'm left with a branch that has updates to the `build/griddle.js` as well as including the `modules` folder, so I'm not sure how to proceed with a PR.

[...]

PR #295, opened 2015-12-21
 Add customRowExtraData prop to pass in extra data through Griddle component to every customRowComponent instance

dlong500:
This is a pull request that is detailed in #286.
I have only included the modifications to the source files (no build or modules).  Let me know if there are issues or anything that needs to be changed.

### DIFF
--- a/scripts/griddle.jsx
+++ b/scripts/griddle.jsx
@@ -63,6 +63,7 @@ var Griddle = React.createClass({
             "useGriddleStyles": true,
             "useGriddleIcons": true,
             "customRowComponent": null,
+            "customRowExtraData": null,
             "customGridComponent": null,
             "customPagerComponent": {},
             "customFilterComponent": null,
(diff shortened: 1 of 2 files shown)


ryanlanciaux:
Thanks for doing this and sorry about my delay in getting back with you on the issue you opened too! I will try this out today/tomorrow and hopefully get it merged then :) 

Really appreciated!

dlong500:
Have you had a chance to look at this yet?  I'd be interested if my PR looks OK or if you see issues with it.  It seems to be working fine for me in my project.

scripts/griddle.jsx
```diff
@@ -707,11 +708,11 @@ var Griddle = React.createClass({
                 hasMorePages={hasMorePages}
                 onRowClick={this.props.onRowClick}/></div>)
     },
-    getContentSection: function(data, cols, meta, pagingContent, hasMorePages){
+    getContentSection: function(data, extraData, cols, meta, pagingContent, hasMorePages){
```
ryanlanciaux: Could we please make extraData the last parameter here?

ryanlanciaux:
So sorry about the delay -- I left one kind of nitpicky comment but after that lets get this merged in :+1:

scripts/griddle.jsx
dlong500: Sure, that's fine. I'll update it shortly.

dlong500: Should extraData be the last parameter in all of the function calls or just in `getContentSection`?

dlong500: Also, don't mean to make things more complicated, but after thinking about it further I was wondering if it might make more sense to call the parameter globalData instead of extraData since it represents data that is global to the Griddle instance (doesn't change per row).  If you like that idea I can just create another completely separate PR and we can close this one.

ryanlanciaux: Yea that's a great idea to make it more reusable. Thanks again for your time and thinking around this!

dlong500:
Closed in favor of #302

PR #297, opened 2015-12-22
Revert "Updated Griddle columns to refresh on componentWillReceiveProps when …"

blkmutt:
Reverts GriddleGriddle/Griddle#296


blkmutt:
Undoing my revert.

PR #299, opened 2015-12-30
Fix bug with React peer dependency version

licyeus:
Griddle's throwing an error when trying to install alongside React 0.14.5.


ryanlanciaux:
Thanks!

pk-nb:
Any chance a version 0.3.1 could be published with this change? This is breaking some of our builds with  other dependencies requiring React ^0.14.5.

PR #302, opened 2016-01-03
 Add globalData prop to pass in user-specified global data through Griddle component to every customRowComponent instance.

dlong500:
This is a pull request that is detailed in #286, and a refining of the now closed PR #295.  The `globalData` prop passed into Griddle will be sent in as a prop to each customRowComponent instance allowing the row component to have access to any user-specified global data.  Without this a custom row component is in somewhat of a "black box" and requires inefficient methods to access state that isn't row specific (like having to connect each row to a flux/redux store).  Right now the `globalData` prop is only passed into instances of customRowComponent, but this could potentially be used in other places since the data can be as generic as a developer wants.

### DIFF
--- a/scripts/customRowComponentContainer.jsx
+++ b/scripts/customRowComponentContainer.jsx
@@ -13,7 +13,8 @@ var CustomRowComponentContainer = React.createClass({
       "data": [],
       "metadataColumns": [],
       "className": "",
-      "customComponent": {}
+      "customComponent": {},
+      "globalData": {}
     }
   },
   render: function() {
@@ -25,7 +26,7 @@ var CustomRowComponentContainer = React.createClass({
     }
 
     var nodes = this.props.data.map(function(row, index){
-        return <that.props.customComponent data={row} metadataColumns={that.props.metadataColumns} key={index} />
+        return <that.props.customComponent data={row} metadataColumns={that.props.metadataColumns} key={index} globalData={that.props.globalData} />
     });
 
     var footer = this.props.showPager&&this.props.pagingContent;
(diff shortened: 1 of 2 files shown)


ryanlanciaux:
Thanks for adding this!

dlong500:
No problem.  Any chance we could get a new release now that this has been merged?

ryanlanciaux:
Yea - sorry about the delay. We'll get one published this weekend. Thanks again :+1:

dlong500:
Hey Ryan, don't mean to bug but is there any chance to get a new release with this change before 1.0?  By the way, I'm glad to see a lot of activity on the 1.0 codebase and look forward to how that turns out.

THPubs:
@dlong500 Hi, can you please tell us exactly how to use this feature? Some documentation? Also what's the lowest release which have this feature?

dlong500:
@THPubs I believe the feature is available in v0.4.0 and above.

Basically, all this does is allow you to set a `globalData` prop on the Griddle component that will be passed into each `customRowComponent` instance so you pass in data that otherwise wouldn't be available to the custom row component.  So in the custom row component you can access the data via `this.props.globalData`.

Given all the restructuring going on leading up to a 1.0 release I'll probably wait to try adding a blurb about it in the docs, but it really is just adding that one pass-through prop for custom rows.

THPubs:
@dlong500 Great thanks! Just got it. Hoping to see it soon in the documentation because it is quite a useful feature.

bhishp:
This doesn't look like it is part of Griddle any more? How would I go about passing a custom prop to a row in this case? Can't quite work it out from the docs

David541:
Was this dropped in favor of the extraData? If so, is there any ETA on when that will be implemented/fixed?

Issue #185, opened 2015-07-16
subgrid icons are not displayed if metadata


pierrepinard-2:
Apparently, the component or characters that are added into the first column of a subgrid row (parentRowCollapsedComponent and parentRowExpandedComponent, default or custom) are not displayed when a metadata object (from columnMetadata) exists for this first column.

This is a problem if we want to customize the displayed name of this column, for example.

Thanks!

sameetandpotatoes:
+1. I was able to reproduce this error too

Thermos192:
+1. I just started using Griddle and this issue popped up for me.  Any work arounds available?

brownieboy:
Another big +1.

Before finding this bug report, I had wasted several hours tweaking my code, unable to understand why the subgrid icon wasn't appearing in the table.  Simply removing my columnMetadata props object "fixed" it.   Only now my column headings are showing the underlying field names, which is just ugly.

This needs a fix or workaround urgently.

brownieboy:
For the moment, I've worked around this by using the CSS ::before selector to add some plus and minus characters, like so:

    .griddle-container .parent-row td:first-child::before {
	content: "+ ";
    }
    .griddle-container .parent-row.expanded td:first-child::before {
	content: "- ";
    }

tmartensen:
+1. This is still happening on version 0.3.0

PR #304, opened 2016-01-04
Subgrid Icons Not Displayed if Metadata Present

cmain:
Fix for the following issue:
The component or characters that are added into the first column of a subgrid row
(parentRowCollapsedComponent and parentRowExpandedComponent, default or custom) are
not displayed when a metadata object (from columnMetadata) exists for this first column.


tmartensen:
Nice! This should fix https://github.com/GriddleGriddle/Griddle/issues/185.

Issue #35, opened 2014-12-05
Custom column headers


0xdabbad00:
Similar to issues #32 and #16 I would like to be able to modify the column header. Specifically, I would like to be able to add [React-Bootstrap tooltips](https://react-bootstrap.github.io/components.html#tooltips) to my column headers.  So when you hover over a column header it explains in more detail what that header means.

Related to this is being able to have a column header reference name that is different than the text that is displayed there.  For example, is it possible to have a column named something like `Order Number` but internally it is referred to as `OrderNum`?  Or maybe you make your site international and use the text `Número de orden` for Spanish speakers, but I want to be able to write code that just references a single variable name.  

The place where I'm currently experiencing some amount of nuisance with this, is when external results gets sorted.  I have to convert `Count` to `NumSystems` on my server for the code shown here: https://github.com/DynamicTyped/Griddle/issues/14#issuecomment-64094434

If I then decide to rename it on the front-end to `Number of Systems`, I'll have to make another change on my server.  I would prefer not to have to make changes to my server when I only change some text on my front-end.

This is sort of two issues, but very related as it all comes down to wanting to modify what is shown for my column header, and wanting to be able to reference that column using a different name.

jazayerim:
Hi @ryanlanciaux. Do you have any updates on this issue? I'm gonna render a react component in the header row instead of a string. Do you have any suggestion or advice?

CrossfireCurt:
Also interested in this!

Plummat:
+1 here. would go with custom rows quite nicely. any progress?

ryanlanciaux:
Hey there -- we are making progress on the next version and this is something that will be available. Beta _should_ be out in the next week or two.

ryanlanciaux:
So we totally missed the week or two (rewrote another part of what we were changing). We are moving this issue to the griddle-render project https://github.com/GriddleGriddle/griddle-render/issues/13 and using it as acceptance criteria for when we release the new stuff. Closing this one for now but feel free to comment here or on the other with any questions.

piush-gupta:
@ryanlanciaux , hi, i got stuck in a problem where i am passing a function declared in react component , to another custom component.

following is my column meta for 'Batch' component:

var columnMeta = [
{
	"columnName":"id",
	"displayName":"Select",
	"customComponent":BatchID,
	"customComponentProps": {myFunc : this.getExternalData}
}
]

[...]

PR #311, opened 2016-01-10
implement custom headers

astraw:
This is related to #35 - it adds the ability to specify a custom header component (and to pass arbitrary properties).

There is one thing that is not general, however -- in order to function, it ascends the DOM tree from the header looking for a `<th>` tag, which it assumes is the encapsulating Griddle header tag. This will fail in the event that your custom header component includes its own `<th>` tag.


joellanciaux:
I shifted a few things around to take out `ascendParentsUntil(...)`, but everything else is awesome.

Thanks a ton for putting this together! :+1:

PR #312, opened 2016-01-14
show "No data" section inside table + Uncaught Error: Invariant Violation fixes

jsdmc:


joellanciaux:
Awesome! I _did_ make a small update so that the paging content doesn't display an empty select / next button.

Thanks for putting this together :+1:

PR #331, opened 2016-01-27
Switch from underscore to lodash modules

yorchv:
This is a second try, by submitting a pull request from my forked master branch


ryanlanciaux:
Thanks for adding this -- I'm seeing something on the infinite scrolling page on the docs. If I run `grunt` and go to  http://localhost:9000/docs/html/infiniteScroll.html it doesn't render the data when it's a local grid. Any thoughts on why that would be happening?

yorchv:
Let me check.

yorchv:
The problem is that the code of griddleWithCallback in the examples dir is not getting the last version. How are you building that example page?

ryanlanciaux:
Thanks for checking into this. I'll double check and make sure our docs build isn't using the old version.

yorchv:
Any update on this?

ryanlanciaux:
I see that warning about the withCallback. I don't know that's related (def. a problem but not sure that it is what's causing this not to render) since i get those warnings without this change. Looking into this a bit because I'd like to get this going this weekend. I'll keep you posted with what I find.

ryanlanciaux:
I think I have it. This appears to be one of the cases where the methods between lodash and underscore don't match up ([See this line](https://github.com/GriddleGriddle/Griddle/pull/331/files#diff-0099b13efb6b4662e728a9cccba35161R454)). I changed first here to be lodash.take instead of lodash.first and all seems to work. I'm going to try to rebase this ahead of master and get it merged into master.

Thanks for the excellent work here and so sorry about my delay in getting this merged. Your help here and in the other repos has been *greatly* appreciated!

ryanlanciaux:
Thanks so much! I have merged your changes from the command line and made some slight changes so I'm going to close this PR but your commits are now there :+1:

That'll be nice to use the smaller lodash modules instead of the full underscore library!

PR #339, opened 2016-02-07
Fix for row cells order in case of numeric columns names for Chrome.

bsalex:
Unfortunately I can not reproduce the problem with PhantomJS, so I've added tests for chrome. 
It is related to https://bugs.chromium.org/p/v8/issues/detail?id=164 .


bsalex:
Hi! 
Do you have any comments on this request? Can we merge it?

PR #350, opened 2016-03-03
Remove circular dependency between griddle and gridRowContainer

bouk:
We're using a webpack/browserify-style bundler internally (to be open-sourced at some point) that only supports es6-style requires at the top of the file, so a dynamic require like used here to resolve a circular dependency doesn't work. I changed the code to have griddle inject the class instead.

cc @kevinhughes27


ryanlanciaux:
Thank you!

PR #357, opened 2016-03-12
Add column filters

ryanlanciaux:
Work in progress for filtering just by columns with a rudimentary example. The approach that we're taking is to say "use a custom heading component combined with a built in filterByColumn method". We are going this route because we want to avoid locking people into our styling on that.

### DIFF
--- a/examples/customization/customColumn.html
+++ b/examples/customization/customColumn.html
@@ -11,6 +11,10 @@ var LinkComponent = React.createClass({
 });
 
 var HeaderComponent = React.createClass({
+  textOnClick: function(e) {
+    e.stopPropagation();
+  },
+
   filterText: function(e) {
     this.props.filterByColumn(e.target.value, this.props.columnName)
   },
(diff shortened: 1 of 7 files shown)


examples/customization/customColumn.html
```diff
@@ -11,8 +11,21 @@
 });
 
 var HeaderComponent = React.createClass({
+  textOnClick: function(e) {
```
ryanlanciaux: This is an example in the documentation. Basically we're creating a custom header component to be used as something that will filter the data by the current column.

scripts/griddle.jsx
```diff
@@ -153,6 +154,54 @@ var Griddle = React.createClass({
...
+          this.props.customFilterer(this.props.results, filter) :
+          this.defaultFilter(this.props.results, filter);
+      } else {
+        newState.filteredResults = null;
+      }
+
+      this.setState(newState);
+    },
+
+    filterByColumn: function(filter, column) {
```
ryanlanciaux: Here we're defining the column filters. The `filterByColumnFilters` method takes all the columnFilters and sorts the data by them -- this only sets them and calls the other function.

scripts/griddle.jsx
```diff
@@ -153,6 +154,54 @@ var Griddle = React.createClass({
            return false;
        });
     },
+
+    filterByColumnFilters(columnFilters) {
+      var filteredResults = Object.keys(columnFilters).reduce(function(previous, current) {
```
ryanlanciaux: Here, we reduce the data by the columnFilters. The result of filtering the data by one column is passed into the next iteration of reduce (as previous).

ryanlanciaux:
@joellanciaux | @kevinold -- Could one of you please take a look over this before I merge it? :) 

I've added some gh comments to hopefully help in that regard. We will need to kick off a 0.4.0 npm release soon after getting this merged. :shipit:

PR #365, opened 2016-03-19
Select All Box Bug Fix

thangaduraicse:
When i tried to use isMultipleSelection in griddle. I got select all box error like this this._updateSelectedRowIds not defined. "this" inside the forEach loop points to this loop row only. So, I assigned this to some other variable, and used it.
```
var self = this;
forEach(visibleRows, function (row) {
    self._updateSelectedRowIds(row[self.props.uniqueIdentifier], newSelectedRowIds, newIsSelectAllChecked);
}, this);
```


ryanlanciaux:
Thanks for adding this! :+1:

PR #366, opened 2016-03-23
NoData Component has no colspan

Julusian:
The cell used for the NoData component did not have a colspan applied, meaning that narrow columns would cause the contents to wrap unnecessarily or would not allow for centre aligning the content.

Before:
screenshot from 2016-03-23 00-02-29

After:
screenshot from 2016-03-23 00-25-01


ryanlanciaux:
Thanks for adding this :+1:

PR #367, opened 2016-03-24
columnMetadata is ignored if data does not contain a matching field

nscarcella:
Hi,

I needed to create a table with an extra "actions" column at the end but the cells for that column would not show up unless my data had an "actions" field (it does not matter what it has, though). Now, since I didn't want to map my whole data set I took a peek in the source and what I found was that the columnMetadata seems to be ignored unless there is a field in the row data with the same name.

I made a little update and now this works for me. Please take a look and let me know if it is of any help.

### DIFF
--- a/build/Griddle.js
+++ b/build/Griddle.js
@@ -82,33 +82,33 @@ return /******/ (function(modules) { // webpackBootstrap
 
 	var React = __webpack_require__(2);
 	var GridTable = __webpack_require__(3);
-	var GridFilter = __webpack_require__(26);
-	var GridPagination = __webpack_require__(27);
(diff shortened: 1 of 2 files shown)


prycejs:
Jus writing to add that I am experiencing this same problem since a recent update, and that I expect nscarcella's fix here would help us as well.

pgilad:
I believe this is the problematic line:
https://github.com/GriddleGriddle/Griddle/blob/master/scripts/griddle.jsx#L591

It verifies that the column actually exists as a property in the `rowData`, and is not `undefined` or an empty string. `Griddle` should let the user filter their own data so that check needs to be removed..

Issue #82, opened 2015-02-10
Please create a bower package


JosephShering:
That would be swell for some of us not using npm.

danawoodman:
@JosephShering did you try and install it with bower (`bower install DynamicTyped/Griddle`) and then link to the file `build/griddle.js`?

JosephShering:
Yeah, I get: 

```
bower not-cached    git://github.com/DynamicTyped/Griddle.git#*
bower resolve       git://github.com/DynamicTyped/Griddle.git#*
bower download      https://github.com/DynamicTyped/Griddle/archive/v0.2.0.tar.gz
bower extract       Griddle#* archive.tar.gz
bower invalid-meta  Griddle is missing "main" entry in bower.json
bower invalid-meta  Griddle is missing "ignore" entry in bower.json
bower resolved      git://github.com/DynamicTyped/Griddle.git#0.2.0
bower install       Griddle#0.2.0
```

ryanlanciaux:
Thanks for the suggestion. We will add to bower soon :+1:

danawoodman:
@JosephShering Weird... it's working for me, anything specific about your environment?

emombay:
Any way we could get an update of the bower package? It's currently pointing to v0.2.15

emombay:
Not sure that worked... this is what I get after clearing bower cache:
```
bower griddle-react#>=0     not-cached git://github.com/GriddleGriddle/Griddle.git#>=0
bower griddle-react#>=0        resolve git://github.com/GriddleGriddle/Griddle.git#>=0
bower griddle-react#>=0       download https://github.com/GriddleGriddle/Griddle/archive/v0.2.15.tar.gz
bower griddle-react#>=0        extract archive.tar.gz
bower griddle-react#>=0       mismatch Version declared in the json (0.2.13) is different than the resolved one (0.2.15)
bower griddle-react#>=0       resolved git://github.com/GriddleGriddle/Griddle.git#0.2.15
bower griddle-react#>=0        install griddle-react#0.2.15

griddle-react#0.2.15 bower_components\griddle-react
├── react#0.14.7
└── underscore#1.8.3
```
I do have the latest bower version.

hagabaka:
@ryanlanciaux needs to add the v0.4.1 git tag, using the commands I included in #373. I can't do that in a pull request.

ryanlanciaux:
[...]

PR #373, opened 2016-03-31
Update version in bower.json

hagabaka:
To update the registered bower package, the upstream owner needs to update git tags:

git tag --annotate --force --message='0.4.1' v0.4.1 be3ea136af6219d544f50a61873db0b9687151d6
git push --tags

That should resolve the latest comment in #82.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "griddle-react",
-  "version": "0.2.13",
+  "version": "0.4.1",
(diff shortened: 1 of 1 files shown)


ryanlanciaux:
Thanks so much!

hagabaka:
@ryanlanciaux No problem. You still need to run those git commands to add the 0.4.1 tag though. (It doesn't seem possible to include them in pull requests.)

PR #229, opened 2015-09-13
Support columns with dots

vampolo:
Suppose you have an array of object of data that you want to display

```
let data = [{
  'a.b': 1
},{
  'a.b': 2
}]
```

powerPick function will interpret 'a.b' as a path and will try to access
data.a.b while it should simply access data['a.b'].

This commit fixes this behavior and thus the issue #228

PR #380, opened 2016-04-08
Support columns with dots

### LINKED ISSUE #228: columns with dots in their property names will show empty table
Hi,

Was trying to show a simple grid with this data

```
let data = [{
    "evt.cpu": 2,
    "evt.dir": "<",
    "evt.info": "",
    "evt.num": 57002,
    "evt.reltime": 1721573085,
    "evt.type": "gettid",
},{
    "evt.cpu": 2,
    "evt.dir": "<",
    "evt.info": "",
    "evt.num": 57003,
    "evt.reltime": 1721573086,
    "evt.type": "gettid",
}];
```

configuring griddle as

```
            <Griddle
        results={data}
        showFilter={true}
        showSettings={true}
        enableInfiniteScroll={true}
        useFixedHeader={true}
        bodyHeight={400}/>
 ```

And it shows a white griddle, i can see that the row is there, but the data is not displayed

Changing the data in order not to have dots in the properties fixes the problem although that is not feasible in my case.

Dem0n3D:
This PR fixes #228. It based on #229, but also solves problem with sorting, which uses getAt (getPath).

### DIFF
--- a/scripts/deep.js
+++ b/scripts/deep.js
@@ -31,7 +31,12 @@ function keysFromPath(path) {
 // path described by the keys given. Keys may be given as an array
 // or as a dot-separated string.
 function getPath (obj, ks) {
-  ks = typeof ks == "string" ? keysFromPath(ks) : ks;
+  if (typeof ks == "string") {
+    if(obj[ks] !== undefined) {
+      return obj[ks];
+    }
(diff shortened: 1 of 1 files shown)


ryanlanciaux:
👍

Issue #337, opened 2016-02-02
Custom sorting on a specific column with useExternal set to false


ShankarSangoli:
I think griddle should allow customizing the sorting logic for a specific column if required irrespective of `useExternal` flag.

Sometimes the data is not flat enough that reversing the data will sort it as needed. In such cases we can give the control to consumer to sort and provide the sorted data.

ryanlanciaux:
Hi and sorry about the extreme delay in responding to this issue. We now allow custom sorting options in Griddle. Please see [the API documentation](http://griddlegriddle.github.io/Griddle/docs/api/) under "ColumnDefinition" and "sortMethod" for more information on how to achieve this in Griddle 1.0.

Closing this one for now but please feel free to open a new issue if any problems are encountered. 👍

Issue #290, opened 2015-12-15
How to sort a column numerically, not alphabetically?


pupsanka:
I have a column that displays floating number, and the sort is done alphabetically, rather than numerically. I tried to provide an override function with externalChangeSort, but it didn't work, unless useExternal is set to true, but then it requires overriding all methods with external. Is there any way to implement numeric sort?
thanks

legomind:
The ability to add a custom sorting function to the columnMetadata prop would be really great, but I took a quick glance and didn't see anything like that in griddles source.

ryanlanciaux:
There was a PR for starting similar functionality (date rather than numbers) but it's not in the source right now.

https://github.com/GriddleGriddle/Griddle/pull/149/files

standino:
we have the some problem. Please help to add sort column numerically feature. Thanks very much.

Dem0n3D:
This is "natural sort" problem, take a look to my PR.

ryanlanciaux:
Thank you all for the discussion around this and for the noted PR. We have recently released Griddle 1.0 that has a customSort property on ColumnDefintion. Please [see this bit of code](https://github.com/GriddleGriddle/Griddle/blob/master/stories/index.js#L119) in our story definitions for an example of this and [the customization docs](http://griddlegriddle.github.io/Griddle/docs/customization/). Closing this one for now but please feel free to open a new issue if any problems are encountered. :+1:

Issue #250, opened 2015-10-13
Adding custom sorting for column


jintoppy:
I dont see an option to set a custom sorter function to the column meta data? Is it possible? If not, is this in pipeline?

ragebiswas:
+1 to this. It's a bit inconvenient to switch to external data loading just for custom sorting. Is it possible to hook into a method on click of the header row - that way I can at least reorder the data of griddle and force a re-render?

ryanlanciaux:
Agreed - how we have it currently set up is not super flexible. 1.0 will allow this sort of behavior through plugins.

ragebiswas:
Thanks @ryanlanciaux . Love the simplicity of griddle though, eagerly waiting for 1.0 :+1:

ragebiswas:
One way in which I was able to workaround this problem - noting if it helps someone else. I added some meta information to the column text so that sorting by ascending/descending order works. I also added a custom formatter for that column so that I show just the stuff I want.

ADanilychevJr:
@ragebiswas Could you elaborate more on how you did this? My Custom Components are not sorting correctly, instead they sort on another column when the header is clicked, for some reason.

ragebiswas:
@ADanilychevJr Suppose you want to sort on column foo, but wish to modify the logic to something other than the default alphanumeric sort offered by griddle. What you have to do is:
1. use a custom renderer that shows the data that you want
2. map each value of your column to a "sortable" string/number. 

For example, if you want to sort by a field called status, where the values should be sorted as Open < Active < Closed. The default griddle sorting would give this order: Active < Closed < Open.

To fix this, map each status to an int value (open -> 1, active -> 2, closed -> 3). Use this as the column value in griddle, and for rendering use a custom renderer that renders the strings. Since griddle would sort on the values, you would get the correct sorting order.

ADanilychevJr:
@ragebiswas Thanks so much! That was exactly the solution I was looking for. I didn't realize that columnName was what the table sorted on for custom Components (thought it was the plaintext in your custom render)

ryanlanciaux:
[...]

Issue #142, opened 2015-04-29
Sort by date


arianitu:
How do you sort by date?

EugeneZ:
At the moment, you have to use a customComponent to display the date the way you want, and the dataset should have a Number for that column so that it can be sorted. It's not great.

    var MyComponent = React.createClass({
        render: function(){
            var date1 = new Date(123456),
                date2 = new Date(234567),
                date3 = new Date(345678);
    
            var data = [{date: date1.valueOf() }, { date: date2.valueOf() }, { date: date3.valueOf() }],
                metadata = [{columnName: 'date', customComponent: DateDisplay }];
    
            return <Griddle results={data} columnMetadata={metadata}/>;
        }
    });
    
    var DateDisplay = React.createClass({
        render: function(){
            return new Date(this.props.data).toLocaleString();
        }
    });

Just a simple example, did not test this in the least but I'm doing something similar.

Semigradsky:
> and the dataset should have a Number for that column so that it can be sorted

dataset must be strings in ISO format (`1973-11-29T15:23:43.324Z`)

EugeneZ:
> dataset must be strings in ISO format

Number works fine. It's just a plain javascript equality sort:

    function baseCompareAscending(value, other) {
      if (value !== other) {
        var valIsReflexive = value === value,
            othIsReflexive = other === other;

        if (value > other || !valIsReflexive || (value === undefined && othIsReflexive)) {
          return 1;
        }
        if (value < other || !othIsReflexive || (other === undefined && valIsReflexive)) {
          return -1;
        }
      }
      return 0;
    }

Tarun17:
I am working on it. I will be provided as a sortType option which will allow you to use the Date object instead of a string

damassi:
@Tarun17 - any movement on the date object type used as sort?

Tarun17:
Yes I have it implemented and working. I need to write unit tests for it, before it can be pulled into the master branch.

damassi:
@tarun - this branch? https://github.com/GriddleGriddle/Griddle/tree/columnSorting

idanwe:
Now there is `customCompareFn` option that you can use in two formats [source code](https://github.com/GriddleGriddle/Griddle/blob/master/scripts/griddle.jsx#L558-L571)

```javascript
customCompareFn: function (data) {
  return new Date(data)
}
```
or
```javascript
customerCompareFn: function (data1, data2) {
  if (data1 > data2)
[...]

PR #382, opened 2016-04-08
Add custom column sorting possibility

Dem0n3D:
This PR can be used as solution for #337, #290, #250, #142 and #116.

Example for #142:
```
var columnMeta = [{
  columnName: "col1",
  displayName: "COL1",
  sortable: true,
  compare: function(item) {
    return new Date(item);
  }
}, ...]
```

Example for #116, #290 (with naturalSort from [javascript-natural-sort](https://www.npmjs.com/package/javascript-natural-sort)):
```
var columnMeta = [{
  columnName: "col1",
  displayName: "COL1",
  sortable: true,
  compare2: function(a, b) {
    return naturalSort(a, b);
  }
}, ...]
```

- **compare** is used for underscore-style sortBy method (with 1 argument)
- **compare2** is used for standard JS sort method (with 2 arguments)

### DIFF
--- a/scripts/griddle.jsx
+++ b/scripts/griddle.jsx
@@ -485,13 +485,29 @@ var Griddle = React.createClass({
         var that = this;
             //get the correct page size
             if(this.state.sortColumn !== "" || this.props.initialSort !== ""){
-                var sortProperty = _filter(this.props.columnMetadata, {columnName: this.state.sortColumn});
-                sortProperty = (sortProperty.length > 0 && sortProperty[0].hasOwnProperty("sortProperty") && sortProperty[0]["sortProperty"]) || null;
+                var sortColumn = _filter(this.props.columnMetadata, {columnName: this.state.sortColumn});
+                var sortProperty = sortColumn.length > 0 && sortColumn[0].hasOwnProperty("sortProperty") && sortColumn[0]["sortProperty"] || null;
+                // for underscore-style sortBy method (with 1 argument)
+                var compare = sortColumn.length > 0 && sortColumn[0].hasOwnProperty("compare") && sortColumn[0]["compare"] || null;
+                // for standard JS sort method (with 2 arguments)
+                var compare2 = sortColumn.length > 0 && sortColumn[0].hasOwnProperty("compare2") && sortColumn[0]["compare2"] || null;
+                var column = that.state.sortColumn || that.props.initialSort;
 
-                data = sortBy(data, function(item){
-                    return sortProperty ? deep.getAt( item, that.state.sortColumn||that.props.initialSort )[sortProperty] :
-                        deep.getAt( item, that.state.sortColumn||that.props.initialSort );
-                });
+                if(compare2){
+                    data = data.sort(function (a, b) {
+                        return compare2(deep.getAt(a, column), deep.getAt(b, column));
+                    });
+                } else {
+                    data = sortBy(data, function (item) {
+                        if(sortProperty){
+                            return deep.getAt(item, column)[sortProperty];
+                        } else if(compare){
+                            return compare(deep.getAt(item, column));
+                        } else {
+                            return deep.getAt(item, column);
+                        }
+                    });
+                }
 
                 if(this.state.sortAscending === false){
                     data.reverse();


Dem0n3D:
Numerical sort for #116 can be done even simpler:
```
var columnMeta = [{
  columnName: "col1",
  displayName: "COL1",
  sortable: true,
  compare: function(item) {
    return parseFloat(item);
  }
}, ...]
```

rahul1059:
is this merged?
What's the use of sortable attribute in Griddle?

ryanlanciaux:
Thanks for adding this -- useful to have more options on the sort like this.

Dem0n3D:
@rahul1059 "[sortable](http://griddlegriddle.github.io/Griddle/customization.html)" Determines whether or not the user can sort this column (defaults to true, so specify false to disable sort)

shikhagupta2424:
@Dem0n3D  Thanks, Do you have any idea about release date/version for merged code?

ryanlanciaux:
@shikhagupta2424 : I have [one PR](https://github.com/GriddleGriddle/Griddle/pull/386) open to allow React 15 and then we'll do a release. It's currently failing so once I get that straightened out we should be good to go.

@Dem0n3D: Thanks again -- these PRs are awesome and a huge benefit to the repo!

Dem0n3D:
I added PR with documentation #387.

eevah:
@ryanlanciaux, is this new functionality part of griddle now? I installed griddle yesterday, but 'compare' functionality is not working for me. Perhaps it's not available yet?

anakinjay:
Is this in the main branch?  I also cannot get it to work and need it badly.

scripts/griddle.jsx
```diff
@@ -485,13 +485,29 @@ var Griddle = React.createClass({
...
             if(this.state.sortColumn !== "" || this.props.initialSort !== ""){
-                var sortProperty = _filter(this.props.columnMetadata, {columnName: this.state.sortColumn});
-                sortProperty = (sortProperty.length > 0 && sortProperty[0].hasOwnProperty("sortProperty") && sortProperty[0]["sortProperty"]) || null;
-
-                data = sortBy(data, function(item){
-                    return sortProperty ? deep.getAt( item, that.state.sortColumn||that.props.initialSort )[sortProperty] :
-                        deep.getAt( item, that.state.sortColumn||that.props.initialSort );
-                });
+                var sortColumn = _filter(this.props.columnMetadata, {columnName: this.state.sortColumn});
+                var sortProperty = sortColumn.length > 0 && sortColumn[0].hasOwnProperty("sortProperty") && sortColumn[0]["sortProperty"] || null;
```
iyel: What is it for?

Dem0n3D: I don't know :) I just trying to keep old logic. Looks like it was added in 1098014 by @ryanlanciaux.

Dem0n3D: it seems that you can use it when your data contains objects, for example:
```
var data = [{
        "id": 0,
        "name": "Mayer Leonard",
        "address": {
          "city": "Kapowsin",
          "state": "Hawaii"
        },
        "country": "United Kingdom",
        "company": "Ovolo",
        "favoriteNumber": 7
      }, ... ]
```
But, looks like that it not needed anymore, after f3a3d34. From griddle-test.js:
```
it('sets sort filter correctly', function(){
    expect(grid.state.sortColumn).toEqual("");
    grid.changeSort("address.state");
    expect(grid.state.sortColumn).toEqual("address.state");
  });
```

mplis-jetsetter:
Not sure if this is the right place to mention this, but I noticed that some documentation [was added for this feature in the repo](https://github.com/GriddleGriddle/Griddle/commit/88dabf5a6aa4a378129a13e01f87e9a359e7f97c), but the website [doesn't actually contain those changes](http://griddlegriddle.github.io/Griddle/customization.html). I would never have found this important feature if I didn't stumble across this PR.

SimonSelg:
Same here as @mplis-jetsetter. Adding it to the website would be good.

rosensama:
Just an FYI for those that find this feature via this ticket.   Per the 0.6.0 release notes:
> Changed columnMetadata:
> 
> customCompareFn - (optional) a function comparator
> Breaking changes Previous you had to write compare or compare2 to define your custom compare function, now it is only customCompareFn that based on number of arguments behaves as compare or compare2

PR #386, opened 2016-04-13
Use React 15

ryanlanciaux:
Use React 15 and remove all instances of setProps :shame:


Dem0n3D:
The problem is in grunt-jsxhint, which depends on obsolete package react-tools from react 0.14

ryanlanciaux:
@Dem0n3D - I think this should be all set. Thanks for the pointer on jsx-hint. Removed it and updated the tests that were failing. 

@joellanciaux @blkmutt - Could you please double / triple review this before I merge :)

sushilk1991:
It would be great to get this merge and a new release as well.

Issue #213, opened 2015-08-27
Nested Objects not rendering 


jgarcia99:
I have data coming in like this

``` javascript
var orderSchema = new Schema({
    clubCode : { type: String },
    shopID :  { type: Number },
    externalID : { type: String },
    orderID : { type: String, index: true },
    orderNumber : { type: String },
    type : { type: String },
    serviceCode : { type: String },
    unseen : { type: Boolean, default: true },
    externalStatus : {
        status : { type: String },
        time : { type: Date, default: Date.now }
    },
    externalStatusHistory : [{
        status : { type: String },
        time : { type: Date, default: Date.now }
    }],
    status : {
        status : { type: String },
        time : { type: Date, default: Date.now }
    },
    statusHistory : [{
        status : { type: String },
        time : { type: Date, default: Date.now }
    }],
    member : {
        memberNumber : { type: String },
        firstName : { type: String },
        lastName : { type: String },
        phoneNumber : { type: String },
        verified : { type: Boolean, default: false }
    },
    vehicle : {
        year : { type: String },
        make : { type: String },
        model : { type: String },
        trim : { type: String },
        color : { type: String },
        tagNumber : { type: String },
        tagState : { type: String },
        vin : { type: String },
        odometer : { type: Number }
    }
});
```

but griddle only renders clubID, orderID etc not vehicle.year or status.type.

jgarcia99:
```javascript
var Table = React.createClass({

    propTypes: {
        orders: React.PropTypes.arrayOf(React.PropTypes.object).isRequired
    },

    getDefaultProps: function() {
        return {
            orders: []
        };
    },

    render: function() {

        var orders = '';
        if (this.props.orders) {
            orders = this.props.orders.map(function(order, index) {
                return <Row order={order}/>;
            });
        }
        var data = this.props.orders;

        var columnMeta = [
            {
                "columnName": "status.time",
                "displayName": "Date/Time",
                "order": 0,
                "visible": true
            },
            {
                "columnName": "memberName.firstName",
                "displayName": "Name",
                "order": 1,
                "visible": true
            },
            {
                "columnName": "vehicle.make",
                "displayName": "Vehicle",
                "order": 2,
                "visible": true
            },
            {
                "columnName": "orderID",
                "displayName": "Order ID",
                "order": 3,
                "visible": true
            },
            {
                "columnName": "type",
                "displayName": "Type",
                "order": 4,
                "visible": true
            },
            {
[...]

PR #392, opened 2016-04-22
Issue resolved for displaying nested data. (Issue No 213)

atapatel:
- Change in code for detecting nested keys from object.

`
teratee = function(key, obj) { return key in obj; }; 

'
can not detect nested keys from object so i have tweak those code.

For Reference. 
- [#213](https://github.com/GriddleGriddle/Griddle/issues/213)

### DIFF
--- a/scripts/deep.js
+++ b/scripts/deep.js
@@ -54,7 +54,14 @@ function getPath (obj, ks) {
 // Credit: https://github.com/jashkenas/underscore/blob/master/underscore.js
 function powerPick (object, keys) {
   var result = {}, obj = object, iteratee;
-  iteratee = function(key, obj) { return key in obj; };
+  iteratee = function(key, obj) {
+    return key.split(".").every(function(x) {
+        if(typeof obj != "object" || obj === null || ! x in obj)
+            return false;
+        obj = obj[x];
+        return true;
+    });
+  }
 
   obj = Object(obj);
 


iisa:
hi @atapatel , thanks for finding and making this fix!  Q: does this work with deeply nested objects?
ex: top.middle.bottom = "hello"

atapatel:
Yes. It can work with N level of nested set

alexchiri:
Would be great to have this released. 

In the meantime, I tried using the waveRFID/Griddle repo as npm dependency and then npm install, but for some reason the griddle-react folder in node_modules doesn't contain the modules folder (like it would when using the npm package), therefore it cannot find and use griddle-react. Anything else I should do in order to be able to use the waveRFID/Griddle version in my project?

atapatel:
@alexchiri well i haven't published my package on npm. for that may be i need to publish those package.

alexchiri:
@atapatel well, theoretically you can set a github repo as a dependency, instead of the version. Something like this:

`"dependencies": {
    "griddle-react": "waveRFID/Griddle", 
....`

And afterwards if you run `npm install` in your project, it should pull all the files and should be usable like a package from npm. But in this case it isn't. So I was wondering what else should I do to get it running in my project using your repo.

alexchiri:
@atapatel managed to get it running in my project. Did you verify it with nested data with arrays?
For example top[].middle.bottom? 

In my situation I have some columns that have column name: top[0].middle and some that show data outside of the top array. The ones from array are not shown and the griddle table is pretty broken. It used to work in version 0.3.1. Still broken with the version from this pull request...

atapatel:
@alexchiri i haven't done code for your scenario. :(

atapatel:
@alexchiri if you have any solutions than please feel free to share with me.

alexchiri:
@atapatel I'll have a look and see if I can fix it starting from your code. Is this where all the magic is happening? https://github.com/GriddleGriddle/Griddle/pull/392/files#diff-486930079516b14efdf7e2572b02c647R81

scripts/deep.js
```diff
@@ -54,7 +54,14 @@ function getPath (obj, ks) {
 // Credit: https://github.com/jashkenas/underscore/blob/master/underscore.js
 function powerPick (object, keys) {
   var result = {}, obj = object, iteratee;
-  iteratee = function(key, obj) { return key in obj; };
```
atapatel: @alexchiri i have changed here for nested data display

alexchiri: ok, I think I found a solution that handles all kind of nested data, but uses an extra module.

so I replaced powerPick with this:
```
function powerPick(object, keys) {
  var result = {};

  for (var i = 0, length = keys.length; i < length; i++) {
    var key = keys[i];
    result[key] = resolvePath(object, key);
  }

  return result;
}
```

where `resolvePath` is 
```
var resolvePath = require('object-resolve-path');
```

and `object-resolve-path` is a package that resolves all kind of values of nested objects, identified by a string key: https://github.com/capaj/object-resolve-path

does this work for you? how can I help to get this out?

tBaxter:
I would also like to see this merged and released.

ADumaine:
I think this also fixes #423.  

I just ran into this issue after upgrading to 0.5.1.  My solution involved flattening (using Hoek.transform) the results before sending them to griddle and removing all the dot notation references in the  columnNames in the colMetadata and Griddle component. Would much prefer support for the nested objects as in 0.3.x.

Any projection on when this will be merged and released?

blueboxjesse:
Would love updates on this one.

PR #418, opened 2016-05-17
Feature/column sort enhancements

iyel:
**Add**
columnMetadata:
* sortDirectionCycle - (optional) an array of sort directions, specifying the order that they should cycle through as the user repeatedly clicks on the column heading. The default is ```[null, 'asc', 'desc']```

**Changed**
columnMetadata:
* customCompareFn - (optional) a function comparator

**Breaking changes**
Previous you had to write ```compare``` or ```compare2``` to define your custom compare function, now it is only ```customCompareFn``` that based on number of arguments behaves as ```compare``` or ```compare2```


pankajJS:
How to pass rowData to customCompareFn  ?

PR #424, opened 2016-05-30
fix gridRow defaultValue logic

odk211:
`fromPairs` does not work.
https://github.com/GriddleGriddle/Griddle/commit/b65bfe8460848f6e3a1a7942c4b1c3ae3c9f4255#diff-486930079516b14efdf7e2572b02c647R83

### DIFF
--- a/build/Griddle.js
+++ b/build/Griddle.js
@@ -82,37 +82,37 @@ return /******/ (function(modules) { // webpackBootstrap
 
 	var React = __webpack_require__(2);
 	var GridTable = __webpack_require__(3);
-	var GridFilter = __webpack_require__(155);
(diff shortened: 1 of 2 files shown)


odk211:
I want to use griddle 0.5.0 but this regression prevent from it.
Could you check it ? and I hope this patch is released.

ryanlanciaux:
Awesome -- thanks so much for adding this. I'll try to test this out tonight

odk211:
@ryanlanciaux Thanks! It's not a hurry, I use my fork now.
When the latest is released, I use :)

odk211:
rebased, but I'm not sure why `build/Griddle.js` diff is so huge :(

PR #428, opened 2016-06-06
[pagination] add custom pager options

tony-kerz:
allow passing options to custom pager component

### DIFF
--- a/docs/src/markdown/customization.md
+++ b/docs/src/markdown/customization.md
@@ -514,6 +514,11 @@ If you want to customize the paging component, just set the property 'useCustomP
   <dd><strong>object</strong> - The custom pagination component. default: {}</dd>
 </dl>
 
+<dl>
+  <dt>customPagerComponentOptions</dt>
+  <dd><strong>object</strong> - Any options to be passed to the custom pagination component. default: {}</dd>
+</dl>
+
 #####Example#####
 
 ```javascript
--- a/scripts/customPaginationContainer.jsx
+++ b/scripts/customPaginationContainer.jsx
@@ -14,7 +14,8 @@ var CustomPaginationContainer = React.createClass({
       "nextText": "",
       "previousText": "",
       "currentPage": 0,
-      "customPagerComponent": {}
+      "customPagerComponent": {},
+      "customPagerComponentOptions": {}
     }
   },
   render: function() {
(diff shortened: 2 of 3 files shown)


ryanlanciaux:
Thanks!

tony-kerz:
thanks @ryanlanciaux!

i just realized that i neglected to include the results of a build in the `build` folder 😢 

i was going to submit a follow on p/r to remediate, 
but then i realized that the build was creating `build/Griddle.js` (with an upper-case `G`),
and i recalled the trouble we had recently with `build/griddle.js` v `build/Griddle.js`.

i checked your most recent release of the `1.0.0-alpha.x` series and that had `build/griddle.js`,
but `master` has `build/Griddle.js`, so i am confused.

btw, i am building on a macbook/osx if it matters, (i'm guessing some of the case funny-business has to do with cross dev on windows...)

so, should i leave it as is with the build result as `build/Griddle.js` or should i try and tweak to get `build/griddle.js` (i'm thinking this is somewhere in the grunt script)...?

regards,
tony.

tony-kerz:
so... what i really wanted to do was refer to the change branch directly until it got baked into a release, more specifically use something like this in my `package.json`:
```
"dependencies": {
    "griddle-react": "https://github.com/GriddleGriddle/Griddle.git#f1759b3"
```
but Griddle's `package.json` specifies [this](https://github.com/GriddleGriddle/Griddle/blob/bf428b11388fc555bddb827c906db1b4c73992a5/package.json#L12):
```
  "main": "modules/griddle.jsx.js",
```
and Griddle's `.gitignore` specifies [this](https://github.com/GriddleGriddle/Griddle/blob/bf428b11388fc555bddb827c906db1b4c73992a5/.gitignore#L28):
```
modules
```
effectively dis-allowing npm references to specific branches.

so... i guess it turns out the `build` folder isn't really being used,
or maybe some people that don't use webpack use that pre-built file?

or i guess it depends on how clients reference into `node_modules`,
i tend to follow the pattern of relying on whatever is specified in `main` of `package.json`, 
so something like (in es6):

```
import Griddle from 'griddle-react'
```

how would you feel about removing `modules` from `.gitignore` (so that `modules

ryanlanciaux:
I've been thinking that over for a little while now, tbh. I it may be nice to make the "build" a bit nicer. Thinking either use solely the build package for the folks using Griddle via bower and for npm as a umd package or something along those lines. As far as I understand, the build is just for folks on bower (I don't use bower so I cannot really speak to this but that was the intent :|)

On the other hand, it may be cleaner in this pass to just check in the modules folder as you mentioned. I think I'm pretty open to either and agree that it would make it a lot nicer to reference a branch via git. If you do want to PR removing 'modules' from gitignore :+1:

tony-kerz:
thanks ryan! #430

Issue #66, opened 2015-02-02
Add 'filterable' parameter to columnMetaData


nlien:
Hi! First off; thanks for this library- and its ease of use. Nice to see good documentation, especially with lots of examples.

I see that the 'sortable' parameter for columnMetaData has gotten into the code, but it's still not documented.

I'd love to see a boolean parameter (t.ex. 'filterable'), which excludes/includes that particular column when performing filtering.

nlien:
Is this something you are planning to implement?

ryanlanciaux:
Hey @nlien Thank you for the kind words! This is a fantastic idea and is something we plan on implementing in the near future.

mhuggins:
+1, would love to see this!

tony-kerz:
i would love to see a variation of this where a `filterable` property set to `true` for a column would provide a column specific filter input, something similar to this: http://allenfang.github.io/react-bootstrap-table/example.html#filter

ryanlanciaux:
Cool - we are planning on supporting this type of functional as a plugin after 1.0 ships :) If anyone wants to try their hand at it beforehand, please ping me and I'll help you get setup on that :)

karthikiyengar:
I would love to take a crack at this if this is still relevant (and if it falls within my capabilities :smile:). Please let me know if there is anything I should know before I get started on this.

ryanlanciaux:
That would be awesome. I _think_ the operation should mostly be in here https://github.com/GriddleGriddle/Griddle/blob/master/scripts/griddle.jsx#L148. I think right before checking the column against the filter string, we would want to make sure that the column is not `filterable={false}`. Please let me know if that helps.

karthikiyengar:
Thanks for the pointer, Ryan - certainly was helpful. Please let me know my changes are not adequate.

sww314:
Closed by #436

Issue #30, opened 2014-11-27
Better Docs site


ryanlanciaux:
The docs site should contain more examples and how-to information. Additionally, it shouldn't necessarily be entirely on one page.

PR #436, opened 2016-06-17
#66, #30 - Support filterable property in columnMetadata

karthikiyengar:
Please let me know if there is anything else to be done here and I would be happy to follow up.


ryanlanciaux:
Thanks for adding this :tada:

Issue #398, opened 2016-04-29
Rows are not rendering after upgrading from 0.3.1 to 0.5


adamgruber:
I think the switch from underscore to lodash may have caused some issues. For example, in `gridRow.jsx` I noticed in the render method [line 81](https://github.com/GriddleGriddle/Griddle/blob/master/scripts/gridRow.jsx#L81) `_.object` was changed to `fromPairs`. These methods do not work the same and so it causes issues like the one below:

``` js
// If columns is flat array ['one', 'two', 'three']
fromPairs(columns, []) // => { o: "n", t: "h" }

// Expected output is 
{
  'one': undefined,
  'two': undefined,
  'three': undefined
}
```

I think [zipObject](https://lodash.com/docs#zipObject) method would work instead

ryanlanciaux:
Thanks for reporting this -- any way to get a PR to fix :smile: :+1:

adamgruber:
Is `columns` always expected to be a flat array or are there times when it is something else?

atapatel:
@adamgruber i don't think so zipObject will work in this situations.
@ryanlanciaux hey i try to fix those nested data display issue. can u please have a look. https://github.com/GriddleGriddle/Griddle/pull/392

atapatel:
@ryanlanciaux sorry to bother you :(

adamgruber:
@atapatel Can you give some more details why you think it will not work? Maybe an example?

atapatel:
@adamgruber Well instead of zipObject we can use transform one. 

see this snippet here. http://codepen.io/ajaypatel/pen/eZQLjv

adamgruber:
@atapatel From what I can see, `columns` should always be a flat array of strings. So, `zipObject` will work just fine

http://codepen.io/adamgruber/pen/yOGVbX?editors=0012

anaximander:
Just opened a PR to fix this... I think there was a bit of a communication breakdown between @adamgruber and @atapatel.  When @atapatel said "i don't think so zipObject will work in this situations.", he meant "I don't think so.  zipObject will work in this situation" which appeared to have been interpreted by @adamgruber as "I don't think zipObject will work in this situation".  

Let me know if there are any concerns with the PR, happy to update it as needed.

adamgruber:
@anaximander Thanks for the PR and clarification. Glad this was resolved.

PR #446, opened 2016-06-29
Using lodash function that provides correct functionality

anaximander:
This is a tiny PR to fix issue #398.  This bug is blocking me (and others) from upgrading to 0.6.0 and using the latest features.


ryanlanciaux:
Thanks for adding this!

PR #455, opened 2016-07-13
Overcome "Empty table header" accessibility error

liorbentov:
See http://webaim.org/techniques/tables/data

### DIFF
--- a/scripts/gridTitle.jsx
+++ b/scripts/gridTitle.jsx
@@ -88,13 +88,15 @@ var GridTitle = React.createClass({
                 }
             }
             titleStyles = meta && meta.titleStyles ? assign({}, defaultTitleStyles, meta.titleStyles) : assign({}, defaultTitleStyles);
+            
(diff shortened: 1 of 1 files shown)


liorbentov:
@ryanlanciaux, can you take a look?

ryanlanciaux:
Sorry about the delay -- Haven't had much time to devote to this yet (been traveling). Will try to soon! Thanks for adding this.

ryanlanciaux:
I'd like to be better at accessibility and I am not sure I fully understand. Is the main reasoning here that we want to display a td if there is no text for accessibility?

liorbentov:
@ryanlanciaux, Yes, we need to display td when the th is empty.
Please take a look at https://www.w3.org/TR/WCAG20-TECHS/F46.html

ryanlanciaux:
Awesome - thanks for the info and for creating this PR. 👍

PR #456, opened 2016-07-19
Format booleans correctly

megawac:
Currently booleans are rendered as an empty cell

This change causes them to be coerced to string before passing to react to render

-----------------------------------

I think the next step to this pr would be to add a `DefaultCell` component (i.e. just the span creation logic) with a formatter


ryanlanciaux:
Great! Thank you!

orangecoding:
👍 nice .

PR #462, opened 2016-08-11
onRowMouseEnter and onRowMouseLeave Griddle events

goooseman:
The `onRowMouseEnter` and `onRowMouseLeave` props can be passed to `Griddle` component.
This props should contain a function, which will be triggered by onMouseEnter and onMouseLeave events of each row with data.

I have added a functionality and added new props to docs.

UPD: `onRowWillMount` and `onRowWillUnmount` were added

### DIFF
--- a/build/Griddle.js
+++ b/build/Griddle.js
@@ -82,37 +82,37 @@ return /******/ (function(modules) { // webpackBootstrap
 
 	var React = __webpack_require__(2);
 	var GridTable = __webpack_require__(3);
-	var GridFilter = __webpack_require__(154);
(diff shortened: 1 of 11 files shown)


vstlouis:
Any chance this is going to get merged in? 😄

ryanlanciaux:
Thanks for adding this! :+1:

blunatic:
I'm not able to get `onRowMouseEnter` working in `v0.8.1` and I'm trying to determine how the prop gets passed down from griddle.jsx.js to gridRow.jsx.js, but I'm not quite seeing it. Any ideas?

PR #484, opened 2016-10-07
Fix 'Rows per page' value problem for npm users

tatethurston:
Commit 26b44f sought to fix the external source 'rows per page' value
problem but did not update the modules/griddle.jsx.js, which is required
for users who install Griddle via npm.


ryanlanciaux:
Thanks a ton

PR #486, opened 2016-10-07
Fix sorting direction icon when useExternal is set

### LINKED ISSUE #457: Sorting direction icon doesn't show up when useExternal is set
Using Griddle, I'm loading data externally by setting useExternal to true, and I found that it doesn't render the sorting direction icon when header cell is clicked for sorting.
The data sorting works with externalChangeSort but the icon doesn't just show up.

I found some related code on line 80 in /modules/gridTitle.jsx.js
```
if (that.props.sortSettings.sortColumn == col && that.props.sortSettings.sortDirection === 'asc') {
                columnSort = that.props.sortSettings.sortAscendingClassName;

```
Here the sortDirection is always null when useExternal is set to true.

And inside changeSort function in /modules/griddle.jsx.js, it won't set state of sortDirection simply returning when useExternal is set.
```
if (this.props.useExternal) {
    this.props.externalChangeSort(column, this.props.externalSortColumn === column ? !this.props.externalSortAscending : true);
    return;
}
```
Looking forward to have a quick fix on this.

tatethurston:
Fixes #457.

### DIFF
--- a/build/Griddle.js
+++ b/build/Griddle.js
@@ -411,10 +411,15 @@ return /******/ (function(modules) { // webpackBootstrap
 	            return;
 	        }
 
-	        if (this.props.useExternal) {
-	            this.props.externalChangeSort(column, this.props.externalSortColumn === column ? !this.props.externalSortAscending : true);
-	            return;
-	        }
+               if (this.props.useExternal) {
+                   var isAscending = this.props.externalSortColumn === column ? !this.props.externalSortAscending : true;
+                   this.setState({
+                     sortColumn: column,
+                     sortDirection: isAscending ? 'asc' : 'desc' 
+                   });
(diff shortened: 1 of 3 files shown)


tatethurston:
rebuilding

ryanlanciaux:
Awesome! Thank you! 👍

jay1302:
In which version can we expect this change?

tatethurston:
@jay1302 I can't speak to versioning, but if you want to use this now just replace the current entry for `griddle-react` in your `package.json` with `"griddle-react": "git://github.com/GriddleGriddle/Griddle.git#b57372c"`

jay1302:
Its working. Thank you!

jay1302:
Wen can we expect fix for fixed header without infinite scrolling?

PR #497, opened 2016-10-21
Column sort

ryanlanciaux:
This adds column sorting back in. This currently is only setup for one column. The intention is that adding additional column sorting will be through a plugin (should be able to override the container and add more objects to the columnSort array).


src/components/TableHeading.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
-const TableHeading = ({ columnTitles, TableHeadingCell }) => {
-  const headingCells = columnTitles.map(t => <TableHeadingCell key={t} title={t} />)
+const TableHeading = ({ columnTitles, columnIds, TableHeadingCell }) => {
```
ryanlanciaux: Iterate over the columns and pass in the column name and the columnId

src/plugins/local/components/TableHeadingCellContainer.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { compose, withHandlers, mapProps } from 'recompose';
+import { sortPropertyByIdSelector } from '../../../selectors/dataSelectors';
+import { setSortColumn } from '../../../actions';
+
+function setSortProperties({setSortColumn, sortProperty, columnId}) {
```
ryanlanciaux: Will eventually have jsdoc but this basically wraps the event in a closure and sets up the sortProperty object for this column and passes it along. The sort property objects are what makes multiSort / multiDirection (aka name could be sort ascending as the primary sort and state could be secondary sort descending -- this is not totally reflected in this PR but keeping the sortProperties setup this way makes that a lot easier).

PR #501, opened 2016-10-25
feature request: allow a custom componentDidMount() method to be passed

jblossom-denver:
this would be useful, in case one needs to interact with the DOM after the Griddle table is rendered

### DIFF
--- a/modules/griddle.jsx.js
+++ b/modules/griddle.jsx.js
@@ -463,6 +463,11 @@ var Griddle = React.createClass({
             });
         }
     },
+    componentDidMount: function componentDidMount() {
+        if(this.props.componentDidMount && typeof this.props.componentDidMount === "function") {
+            return this.props.componentDidMount();
+        }
+    },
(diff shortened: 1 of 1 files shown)


ryanlanciaux:
Thanks for adding this :+1:

jblossom-denver:
@ryanlanciaux no problem, thanks for merging!

jblossom-denver:
@ryanlanciaux will this go into a release and published to npm any time soon? thanks

ryanlanciaux:
So sorry about the delay -- it should be in the next couple of days.. trying to determine if there is anything that belongs in this release or if it'd be better to just push this out and follow up with another point release a little later on.

jblossom-denver:
@ryanlanciaux no worries, thanks for the update. If you ship before I do, I won't have to manage a fork. So I can sit tight for a few days if needed

jblossom-denver:
@ryanlanciaux any updates? thanks :)

PR #509, opened 2016-11-10
Allow date and generate selected columns options from columnMetadata if it exists

babenzele:
Griddle was ignoring my `Date` constructor in `this.props.results`. 

Also, if a field was missing in the first item in `this.props.results`, Griddle was not showing that column in grid settings. 

Please let me know if you have any issues or questions.


ryanlanciaux:
Thanks for adding this -- it looks like a great fix!

babenzele:
Thanks!

PR #515, opened 2016-11-24
Fix sort when using external

flamerohr:
Hi, when I'm using external data Griddle changes the initial sort order that I had given it (asc to desc or vice versa), and this only happens on load.

This fix is to stop it changing sort order on load (changeSort shouldn't be called on init/mount anyway?)


ryanlanciaux:
Thanks for adding this! 👍

jay1302:
@ryanlanciaux Any idea on when next version of griddle will release?

PR #521, opened 2016-12-02
feature request: allow a custom componentDidUpdate() method to be passed

jblossom-denver:
similar to https://github.com/GriddleGriddle/Griddle/pull/501

this would be useful/flexible, in case one needs to interact with the DOM upon the Griddle component's state change, and is not using Redux


ryanlanciaux:
Nice - this a great thought to have here. 👍

jblossom-denver:
@babenzele I was not aware of this

PR #522, opened 2016-12-04
Remove unnecessary replacement of filteredColumns

babenzele:
The removed statement overwrites filteredColumns with the columns prop, preventing filteredColumns from being controlled by column settings. This PR allows filteredColumns to be dynamic and controlled externally via state.


ryanlanciaux:
Thanks for addressing this!

Issue #490, opened 2016-10-11
Column filtering can't handle types other than string


m4m4m4:
Column filtering in PR #357 cant handle types other than string
For instance; the "filterByColumnFilters" function tries to toString() a null type which obviously fails.

It would be great if you can override the column filter function with your own as it works in the general filter, i don't see that it's possible right now.

ryanlanciaux:
This should now be possible with Griddle 1.0. Please [see the customization options for more information](http://griddlegriddle.github.io/Griddle/docs/customization/) on this one.

I'm going to close this one for now but please feel free to open a new issue if any problems are encountered. 👍

PR #524, opened 2016-12-06
columnFilter to filter nested data

### LINKED ISSUE #65: Make filter queries search throw nested divs for custom components
Currently, if you're using a custom row component the filter will not match the text inside nested elements. It would be nice if we could match text a few levels down. Happy to submit a PR for this.

tarnacious:
This fixes #65 "[make filter queries search through nested divs for custom components](https://github.com/GriddleGriddle/Griddle/issues/65)" which is tagged an [approved issue](https://github.com/GriddleGriddle/Griddle/issues?q=is%3Aopen+label%3Aapproved+no%3Aassignee). I needed exactly this myself, so I implemented it.

The basic feature is; given data like this:

    data = [
       {
          somethingElse: [{ blah: "one",
             blah2: ["two"] }],
          ...
      },
      ...
    ]

The column filters can now match "one" and "two".

I have some concerns about how this would perform on large objects, but I also think traversing the objects is a pretty good default. This could be implemented a little more efficient for some cases but in a vast majority of cases it should be fine.   

I will implement custom column filters functions on top of this commit ([as requested here](https://github.com/GriddleGriddle/Griddle/issues/490)) as I need them myself, I will also create a pull request for if this is accepted.

I think the feature works quite well and have added tests. If there are any minor changes desired to get this merged in please let me know, thanks.

### DIFF
--- a/build/Griddle.js
+++ b/build/Griddle.js
@@ -88,30 +88,30 @@ return /******/ (function(modules) { // webpackBootstrap
 	var GridNoData = __webpack_require__(179);
 	var GridRow = __webpack_require__(180);
 	var GridRowContainer = __webpack_require__(162);
-	var CustomRowComponentContainer = __webpack_require__(197);
-	var CustomPaginationContainer = __webpack_require__(198);
-	var CustomFilterContainer = __webpack_require__(199);
+	var CustomRowComponentContainer = __webpack_require__(200);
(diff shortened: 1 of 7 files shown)


joellanciaux:
Merged :+1:

I had to make a _tiny_ merge update, but everything else looks great. Thanks a ton for putting this together!

PR #527, opened 2016-12-08
Stop sorting internally if using external sort

flamerohr:
Had some issues where the internal sorting wasn't agreeing with external sort order given, I could've patched this with a custom sort function... but I think this is a cleaner approach. (and less code duplication for when my sorting happens on the server)

### DIFF
--- a/build/Griddle.js
+++ b/build/Griddle.js
@@ -587,73 +587,72 @@ return /******/ (function(modules) { // webpackBootstrap
 
 	        var that = this;
 
-	        // get the correct page size
-	        if (this.state.sortColumn !== "") {
-	            var column = this.state.sortColumn;
-	            var sortColumn = _filter(this.props.columnMetadata, { columnName: column });
-	            var customCompareFn;
-	            var multiSort = {
-	                columns: [],
-	                orders: []
-	            };
+	        if (!this.props.useExternal) {
+	            if (this.state.sortColumn !== "") {
+	                var column = this.state.sortColumn;
+	                var sortColumn = _filter(this.props.columnMetadata, { columnName: column });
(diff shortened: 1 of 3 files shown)


tractorcow:
As an alternative to this, I would love it if it were possible to access the root objects being sorted in customCompareFn.

E.g.

`return customCompareFn(a, b);`

vs current 

`return customCompareFn(_get(a, column), _get(b, column));`

Then we could do somehing like

```json
{
	columnName: size,
	customCompareFn: (a, b) => ((a.size * a.unitModifier) - (b.size * b.modifier));
}
```

tractorcow:
I found a work-around for @flamerohr 's use case. You can use the below to disable re-sorting of values.

```
{
        columnName: 'title',
        customCompareFn: () => (0), // Disable custom sorting
},
```

flamerohr:
Yes, the workaround is available, but it would require defining for every column, which is not ideal when you've got server-side sorting by use of externalSort, etc.

Access to the row data would be very handy for the customCompare endpoint.

robababa:
+1 for this PR.  Ran into the same issue.  We get data from an external REST API, and when the user changes sorting, we re-fetch data from the REST endpoint, because we use pagination.  Unfortunately, the user sees the old data re-sorted before the new data arrives.  This PR would fix that.

Specifically, when the user clicks a column header to change sorting, it invokes the external function, which creates the action to change the sort parameters in Redux.  The parent container then runs its `componentDidUpdate()` method to request new data from the REST endpoint.  In the meantime, unfortunately, the Griddle component re-sorts the old data, and the user sees this before the new data from the REST endpoint arrives.

PR #562, opened 2017-02-04
Modify module building for new version

StevenTCramer:
I added a small no props story to story book.  (It fails but I think it is a valid story and should at least not fail.)

tried to create a simple webpack config that would work to allow me to npm/yarn link to the griddle package for testing.

All the ava tests passed and also the storybook continued to behave as it did prior to my changes.


ryanlanciaux:
This looks good :+1:  Thanks a ton!

I've removed the additional story around checking for data. This should be a separate PR and ideally have some unit tests around it as well (I realize that's asking a lot since we don't have any unit tests around the components in this branch yet).

PR #564, opened 2017-02-09
Fix object data

StevenTCramer:
It was discovered that when the `data` passed to Griddle was created using either a class like:
```
class personClass {
  constructor(id, name, city, state, country, company, favoriteNumber) {
    this.id = id;
    this.name = name;
    this.city = city;
    this.state = state;
    this.country = country;
    this.company = company;
    this.favoriteNumber = favoriteNumber;
  }
}
```
or via a function like:

```
function person( id, name, city, state, country, company, favoriteNumber)
{
  this.id = id;
  this.name = name;
  this.city = city;
  this.state = state;
  this.country = country;
  this.company = company;
  this.favoriteNumber = favoriteNumber;
}
```
Griddle would fail do to missing `row` function.  This PR just creates test cases (stories) for this.
To switch between data sets there three stories:
* set fakeData to hash Objects
* set fakeData to constructed Objects
* set fakeData to class Objects

### DIFF
--- a/stories/fakeData2.js
+++ b/stories/fakeData2.js
@@ -1,0 +1,29 @@
+import fakeData from './fakeData';
+
+function person( id, name, city, state, country, company, favoriteNumber)
+{
+  this.id = id;
+  this.name = name;
+  this.city = city;
+  this.state = state;
+  this.country = country;
+  this.company = company;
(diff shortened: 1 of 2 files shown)


ryanlanciaux:
Thanks! This has been merged locally with some changes I made.

PR #577, opened 2017-02-20
Fix rowMetaData not updated

### LINKED ISSUE #552: Row metadata never changes unless a key is specified
`rowMetadata` is evaluated every time a `dispatch` is fired from my `onRowClick` event, however as far as `rowMetadata` is concerned, the global state is permanently stuck in its initial state.  So even though I've saved the clicked row data as an object on the global state, when `rowMetadata` is re-evaluated it can't see the updated state.

While googling, I came upon issue #352 and when I specified a `key` property as suggested at the bottom of that thread, `rowMetadata` magically started updating.  I get the feeling that this is not expected since it's not documented anywhere, and even in #352's thread it's mentioned as a band-aid.

In the code below, `stateProps` is what would never be updated.

```
rowMetadata: {
    bodyCssClassName: (rowData) => {
        console.log(stateProps);
        return rowData.requirementId === stateProps.report.requirementId ? 'standard-row selected' : 'standard-row';
    }
}
```

But if I add the `key` property, then everything is fine:

```
<Griddle 
    ...
    rowMetadata={props.rowMetadata}
    key={props.report.requirementId}/>
```

I feel like this is something that's hard to explain, so if I'm doing as bad a job of it as I think I am I can try to provide a working example.

Cerberus:
Fix #552

### DIFF
--- a/scripts/griddle.jsx
+++ b/scripts/griddle.jsx
@@ -390,7 +390,8 @@ var Griddle = React.createClass({
             this.setPageSize(nextProps.resultsPerPage);
         }
 	    //This will updaet the column Metadata
-	    this.columnSettings.columnMetadata = nextProps.columnMetadata;
+        this.columnSettings.columnMetadata = nextProps.columnMetadata;
+	    this.rowSettings.rowMetadata = nextProps.rowMetadata;
         if(nextProps.results.length > 0)
         {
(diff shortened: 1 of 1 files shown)


ryanlanciaux:
Thanks for adding this!

supasate:
Hi @ryanlanciaux,

Is there any plan to release this fix on `npm`?

PR #590, opened 2017-03-01
Add ability to specify row key

joellanciaux:


src/utils/dataUtils.js
```diff
@@ -22,28 +22,28 @@ function fromJSGreedy(js) {
...
-  // build up a new list of data and list lookup
-  for (let i = 0; i < data.length; i += 1) {
-    const key = getKey();
-    lookup[key] = i;
-
-    // get an Immutable map of the data item
-    const map = fromJSGreedy(data[i]);
-    list.push(localSettings.addGriddleKey ? map.set(localSettings.name, key) : map);
-  }
+export function transformData(data, renderProperties) {
```
ryanlanciaux: Does this make `addKeyToCollection` no longer necessary?

src/reducers/dataReducer.js
```diff
@@ -49,7 +48,7 @@ export function GRIDDLE_INITIALIZED(initialState) {
  * @param {Object} action - The action object to work with
 */
 export function GRIDDLE_LOADED_DATA(state, action) {
-  const transformedData = transformData(action.data);
+  const transformedData = transformData(action.data, state.get('renderProperties').toJSON());
```
ryanlanciaux: (I personally like the idea of throwing an error saying something like `given rowKey does not exist` so it's really clear why it's failing)

joellanciaux: We _can_, but I think it'd going to lead to something that's a bit of a bummer during `GRIDDLE_INITIALIZE`, where the initial conversion into Immutable happens.

The alternative to `toJSON()` here would be converting the `renderProperties` in `GRIDDLE_INITIALIZE` to a `Map` before the rest of the initial state and then executing `set('renderProperties', renderPropertiesMap)`.

Both options seem pretty lame, heh.

src/utils/dataUtils.js
joellanciaux: yep -- looks like you yanked it out.

ryanlanciaux: :sunglasses:

src/reducers/dataReducer.js
ryanlanciaux: Got it -- I think this is good then.

PR #594, opened 2017-03-02
Fixes #592 - Custom component in settings

### LINKED ISSUE #592: Custom component disappears from the settings list when hidden
I have a column definition that uses a custom component to display a link. If I hit the Settings button and uncheck the box to display that column, it disappears and does not show up in the hidden columns list as expected. I've attached two screenshots that demonstrate the problem. In the before image, there is a column named "edit" that disappears completely in the after image.

The custom column does reappear if I reload the page.

```
<RowDefinition>
    <ColumnDefinition id="channelName" title="Channel"/>
    <ColumnDefinition id="category" title="Category"/>
     <ColumnDefinition id="status" title="Status"/>
     <ColumnDefinition id="edit" title="Edit" customComponent={enhancedActionColumn}/>
</RowDefinition>
```

before

after

ryanlanciaux:
We were not taking ephemeral columns (columns that are made-up of render properties but do not exist in the data) into account when handling the showing/hiding of columns from settings. This addresses that.

_There still needs to be some tests and organization around the reducer changes before merging this one._ 

This one can be tested by cloning the repository followed by `npm install && npm run storybook`. From there, click on the `Bug fixes` -> `Delete row` story (This was for another potential bug but has the same sort of column). Changing the visibility state of the columns from settings should now work as intended.

### DIFF
--- a/src/reducers/__tests__/dataReducerTest.js
+++ b/src/reducers/__tests__/dataReducerTest.js
@@ -144,3 +144,23 @@ test('toggle column sets true when no columnProperty for column but other column
   const state = reducers.GRIDDLE_TOGGLE_COLUMN(initialState, { columnId: 'state' });
   test.deepEqual(state.getIn(['renderProperties', 'columnProperties', 'state']).toJSON(), { id: 'state', visible: true });
 });
+
+test('toggle column works when there is no visible property', (t) => {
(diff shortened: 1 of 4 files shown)


ryanlanciaux:
cc: @sbanwart :)

Issue #483, opened 2016-09-30
Griddle doesn't refresh when item is added / removed during search


babenzele:
When the user has entered a search term in the search box and they add or remove an item from the results prop, the item doesn't show up / stays in the grid until the user performs another search. 

In the meantime, is there a way to trigger a refresh / redraw of the table?

emorling:
@babenzele how are you handling editing in Griddle? I am somewhat at loss here.

babenzele:
@emorling https://github.com/kaivi/riek

PR #595, opened 2017-03-02
Fix[Filter]: Refresh results during filtering

TooHTooH:
#483 Griddle doesn't refresh when item is added / removed / edited during search


ryanlanciaux:
Thanks a ton for adding this!

TooHTooH:
You're welcome ! :)

PR #610, opened 2017-03-18
Actually map plugin events & settingsComponentObjects

dahlbyk:
## Griddle major version:
1.x

## Why these changes are made
They seem to be typos.

## Are there tests?
Glad to add, if this is indeed the desired behavior.


ryanlanciaux:
Thanks a ton for adding this.

ryanlanciaux:
We need to add more tests around this section in general -- these can be added as a separate PR or they can come with the bigger push to get unit tests around the component registration (we have a bit around the utils that back this up but not so much around the components)

PR #618, opened 2017-03-21
WIP: TypeScript Definition + Stories

dahlbyk:
This should close #611, eventually. I figure the questions asked there are better answered with visuals.

> The other thought is would it be possible to test the types via unit tests instead or would that be way more work? I'd love it if there was something that would allow those who are using JavaScript to continue doing so with Griddle codebase but also make life easier for folks who are using Griddle in a typescript project all while ensuring that the types are up-to-date as we make changes in Griddle.

Ultimately type definitions are about codifying valid usage of the library, so they are best tested with real-world use cases. That could be achieved using a [standalone `tsx` test file](https://github.com/NewBoCo/DefinitelyTyped/blob/b16bfba18fc6c227fba117c4e30fa6b5691ce358/griddle-react/griddle-react-tests.tsx), but I'm not sure if doubling the work would be worth it.

> I do like the thought of validating this through stories but I have some potential concern around converting the storybook over. We are hoping that when bugs are encountered that people will submit a PR with a story that demonstrates the issue. Would converting the storybook over to TypeScript also mean that people would have to use TypeScript for this venture?

I wasn't sure how to quantify how much would need to change, so I just went through with the conversion to demonstrate what needs to be done differently. I'll add a review to this PR to walk through what needed to change, but in short: not much, especially if they can use `fakeData`.

Unfortunately, I apparently don't know Storybook and Webpack well enough to get this over the finish line. I'm getting a rather confusing error that I'm hoping someone can help me troubleshoot:

```
ERROR in ./.storybook/config.js
Module build failed: Error: Could not find file: 'C:\Dev\NewBoCo\Griddle\.storybook\config.js'.
    at getValidSourceFile (C:\Dev\NewBoCo\Griddle\node_modules\typescript\lib\typescript.js:81080:23)
    at Object.getEmitOutput (C:\Dev\NewBoCo\Griddle\node_modules\typescript\lib\typescript.js:81446:30)
    at getEmit (C:\Dev\NewBoCo\Griddle\node_modules\ts-loader\dist\index.js:99:43)
    at Object.loader (C:\Dev\NewBoCo\Griddle\node_modules\ts-loader\dist\index.js:27:11)
```

Finally, I could use some feedback on the the type definition itself. I believe it's relatively complete, except the component types and `GriddleComponents` properties.

**Update:**

* [x] Griddle
  * [ ] GriddleComponents
* [ ] actions
* [ ] components
  * [x] Cell
  * [ ] CellContainer
  * [x] ColumnDefinition
[...]

PR #624, opened 2017-03-30
Table/NoResults Improvements

### LINKED ISSUE #615: When data is empty is it possible to have the table header printed out along with the "No results found." message?
## Griddle version
1.1.0

## Expected Behavior
When data is empty it would be nice if the component printed an empty table with the column headers and a single row displaying an empty message of some sort (eg. No results found.).

## Actual Behavior
A message of "No results found." is returned.

## Steps to reproduce
Provide an empty collection to the griddle component.

dahlbyk:
## Griddle major version

1.x

Acknowledging that #618 is a lot to handle, I figure I might as well pull some of it out (plus some changes) that should be easier to review in isolation.

* A few of the trivial storybook fixes from #618
* A few dev environment improvements:
  * For LF from Git on Windows
  * [cross-env](https://www.npmjs.com/package/cross-env) for builds on Windows
  * [EditorConfig](http://editorconfig.org/) for anyone
* `Table`, `TableHeading` and `NoResults` are now generally more resilient
* `NoResults` props actually work
* If `data` is `null` or `undefined`, `NoResults` is rendered instead of runtime errors (with tests!)
* `NoResults` now receives `columnIds` and `columnTitles` (consistent with `TableHeading`; mostly useful for column count)

## Why these changes are made

The primary use case, for which I added a story, is a `Table` component which renders `NoResults` below its `TableHeading` instead of having `NoResults` replace the whole `<table />`.

Closes #615 

## Are there tests?

Yes!

### DIFF
--- a/stories/index.js
+++ b/stories/index.js
@@ -660,6 +660,18 @@ storiesOf('TableHeading', module)
 
 storiesOf('Table', module)
   .add('base table', () => {
+    const noResults = props => (
+      <p>Nothing!</p>
+    );
+
+    return (
+      <Table
+        NoResults={noResults}
+      />
+    );
+  })
+
+  .add('base table with visibleRows', () => {
     const tableHeading = props => (
       <thead>
         <tr>
@@ -682,6 +694,7 @@ storiesOf('Table', module)
 
     return (
       <Table
+        visibleRows={1}
         TableHeading={tableHeading}
         TableBody={tableBody}
       />
--- a/stories/index.js
+++ b/stories/index.js
@@ -1,5 +1,8 @@
 import React, { Component } from 'react';
 import { storiesOf, action, linkTo } from '@kadira/storybook';
+import compose from 'recompose/compose';
+import mapProps from 'recompose/mapProps';
+import getContext from 'recompose/getContext';
 import withContext from 'recompose/withContext';
 import { connect } from 'react-redux';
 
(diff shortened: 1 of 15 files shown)


dahlbyk:
> Revert "Expose column data to NoResults"

After further consideration, it seems more appropriate for column data not to be included by default, since it's relatively simple to override `NoResultsContainer`. As such, I've reverted that change and updated the story to preserve the desired behavior without modifying the default component.

stories/index.js
```diff
@@ -666,6 +664,69 @@ storiesOf('TableHeading', module)
...
+
+  .add('empty with columns', () => {
+    const components = {
+      Table: ({ TableHeading, TableBody, NoResults, style, visibleRows }) => (
+        <table style={style}>
+          <TableHeading />
+          { visibleRows ? (TableBody && <TableBody />) : (NoResults && <NoResults />) }
+        </table>
+      ),
+      NoResultsContainer: compose(
```
dahlbyk: This seems to work as expected... Is there a reason your `Container` components use the following pattern?

```javascript
const WhateverContainer = OriginalComponent => compose(
  ...
)(props => <OriginalComponent {...props} />);
```

ryanlanciaux: The main reason for making our container components higher-order components rather than wrappers is that we are assuming that they may be composed via plugins, etc. 

There may be a better / clearer way to do this moving forward but that was the initial intent (if I understand correctly :+1:)

dahlbyk: I may be wrong, but my understanding of `compose()` is that it's already returning a higher-order component. The outer arrow function seems to be redundant; `compose` is already returning a function that takes in the original component and returns the composed component.

ryanlanciaux: You're totally right - I'm going to think a little bit about why we're doing this (totally possible it's oversight). Thanks!

ryanlanciaux:
This is excellent thank you and sorry about the delay! :+1:

I'm going to follow-up some more on the arbitrarily nested HoC -- That shouldn't block here.

PR #628, opened 2017-04-02
Settings Component Customization

dahlbyk:
## Griddle major version
1.x

* `components` now exports a `SettingsComponents` key with the default components.
* If `component` is not defined on a `settingsComponentObjects` entry, the corresponding key from `components.SettingsComponents` is used.
  * This is how the default settings components are specified, as this configuration is most flexible.
* If a settings component cannot be found in `settingsComponentObjects` or `components.SettingsComponents`, the setting is ignored (instead of an error).
* A top-level `settingsComponentObjects` prop can now override default/plugin settings components, as with `components` and `renderProperties`.
* Settings components are now ordered by `order`.

Plus a few stories proving out scenarios that seem interesting.

## Why these changes are made

Hooking into `settingsComponentObjects` didn't really feel like the other extensibility points, so I started trying alternatives that might align with what I imagine to be the most common scenarios. Specifically:

1. Removing a settings component (set `components.SettingsComponent.key = null`)
2. Reordering the default settings components, without changing them (provide only `{ order: n }` for the appropriate key(s))
3. Swapping in a replacement component, without reordering (set `components.SettingsComponent.key`)

## Are there tests?

One test; mostly stories showing usage. The `SettingsContainer` changes are most deserving of a test, but there isn't any prior art for testing a container.

### DIFF
--- a/src/settingsComponentObjects/ColumnChooser.js
+++ b/src/settingsComponentObjects/ColumnChooser.js
@@ -40,7 +40,7 @@ return (
             type="checkbox"
             name={visibleColumns[v].id}
             checked
-            onClick={onToggle}
+            onChange={onToggle}
           />
           {visibleColumns[v].title || visibleColumns[v].id}
         </label>
@@ -57,7 +57,7 @@ return (
           <input
             type="checkbox"
             name={hiddenColumns[v].id}
-            onClick={onToggle}
+            onChange={onToggle}
             defaultChecked={false}
           />
           {hiddenColumns[v].title || hiddenColumns[v].id}
--- a/stories/index.js
+++ b/stories/index.js
@@ -862,6 +862,18 @@ storiesOf('Settings', module)
       <Griddle data={fakeData} plugins={[LocalPlugin,plugin]} />
     );
   })
+  .add('reorder built-in settings', () => {
+    const settingsComponentObjects = {
+      before: { order: 1, component: () => <div>Before</div> },
+      columnChooser: { order: 2 },
+      between: { order: 3, component: () => <div>Between</div> },
+      pageSizeSettings: { order: 4 },
+      after: { order: 5, component: () => <div>After</div> },
+    };
+    return (
+      <Griddle data={fakeData} plugins={[LocalPlugin]} settingsComponentObjects={settingsComponentObjects} />
+    );
+  })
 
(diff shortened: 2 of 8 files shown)


src/settingsComponentObjects/ColumnChooser.js
```diff
@@ -40,7 +40,7 @@ return (
             type="checkbox"
             name={visibleColumns[v].id}
             checked
-            onClick={onToggle}
+            onChange={onToggle}
```
ryanlanciaux: Thank you for switching these to the correct events!

ryanlanciaux:
Thanks! This is a great change. "Settings" is one of the areas we gave the least attention and this really helps a lot! 

I can hold off until this is ahead of master -- Also, with the TypeScript PR that is coming -- not sure where this one falls in the ideal merge order. Please let me know but this looks good to go and I'd be more than happy to merge soon -- thank you again for all of these great updates!

dahlbyk:
Do you ha e a preference between merge/rebase to bring PR branches up to date? I am comfortable with either.

Once this is merged, I will definitely rebase the TS branch onto master since it includes some redundant commits.

ryanlanciaux:
Either is good with me also.

stories/index.js
```diff
@@ -16,8 +19,10 @@ import { Table } from '../src/components/Table';
 import TableContainer from '../src/components/TableContainer';
 import ColumnDefinition from '../src/components/ColumnDefinition';
 import RowDefinition from '../src/components/RowDefinition';
+const { SettingsWrapper, SettingsToggle, Settings } = '../src/components';
 import _ from 'lodash';
-import { columnIdsSelector, stylesForComponentSelector } from '../src/selectors/dataSelectors';
+import * as actions from '../src/actions';
+import * as selectors from '../src/selectors/dataSelectors';
```
dahlbyk: In retrospect, I should have merged. Regardless, the conflicts were overlapping import lists here. Rather than merge the lists, it seems reasonable to import the full namespace. This will change again with the import refactoring in #618.

ryanlanciaux: Yea this seems good and sorry if my suggestion of rebasing threw this off.

ryanlanciaux: Thanks a ton for adding this.

Issue #591, opened 2017-03-01
How can I disable sorting for certain columns?


sangnuce:
I've been searching for this for days but I cannot disable sorting for some columns.
Please show me how to do it.
Thank you!

ryanlanciaux:
Could you please note which version you are using? :+1:

sangnuce:
@ryanlanciaux I'm using version 1.0

sangnuce:
@ryanlanciaux please help me!

alexcurtis:
I'd also like to know if this is supported. Using Griddle 1.1.0

hkasemir:
I was able to accomplish this in v1.0.0 by doing the following:

```javascript
function CustomSortTableHeader(props) {
  let {
    // filter out props that don't belong on the th tag
    columnId, customHeadingComponent, icon, setSortColumn,
    sortAscendingIcon, sortDescendingIcon, sortProperty, title,
    ...filteredProps
  } = props;
  if (props.columnId === 'super-special-id') {
    // don't allow sorting on this super special column
    filteredProps.onClick = null;
  }
  return (
    <th {...filteredProps}>
      {title}
    </th>
  );
}

...

<Griddle
  components={{
    TableHeadingCell: CustomSortTableHeader()
  }}>
...
</Griddle>

```

would be interested to know if there is a method that's more built-in, though!

dahlbyk:
Once #653 is shipped, you will be able to set `sortable={false}` on any `ColumnDefinition`.

PR #633, opened 2017-04-06
setSortProperties adjustments + demo of unsortable column

dahlbyk:
## Griddle major version

1.x

1. Two identical default implementations of `setSortProperties` have been replaced with a version in `sortUtils`
2. `LocalPlugin` now respects `events.setSortProperties` before defaulting to the default implementation
3. Functions in the `utils` namespaces are now actually exported
4. Added a story demoing disabling sort on a column, an alternative to @hkasemir's in https://github.com/GriddleGriddle/Griddle/issues/591#issuecomment-292063561
5. Type definitions for `sortUtils`

## Why these changes are made

@hkasemir nerd sniped me:
> would be interested to know if there is a method that's more built-in, though!

I'm curious if anyone can think of a clean extensibility pattern that would allow either the custom `setSortProperties` or Heidi's `CustomSortTableHeader` to access a custom prop on `<ColumnDefinition />`.

## Are there tests?

Just a story 😞

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -12,26 +12,13 @@ import * as selectors from './selectors/dataSelectors';
 import { buildGriddleReducer, buildGriddleComponents } from './utils/compositionUtils';
 import { getColumnProperties } from './utils/columnUtils';
 import { getRowProperties } from './utils/rowUtils';
+import { setSortProperties } from './utils/sortUtils';
 import * as actions from './actions';
 
 const defaultEvents = {
   ...actions,
   onFilter: actions.setFilter,
-  setSortProperties: ({setSortColumn, sortProperty, columnId}) => {
-    return function(event) {
-      if (sortProperty === null) {
-        setSortColumn({ id: columnId, sortAscending: true });
-        return;
-      }
-
-      const newSortProperty = {
-        ...sortProperty,
-        sortAscending: !sortProperty.sortAscending
-      };
-
-      setSortColumn(newSortProperty);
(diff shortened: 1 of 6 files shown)


hkasemir:
For what it's worth, my initial instinct was to override the `onSort` event to provide custom sort behavior, but it also seems unavailable at the moment to provide a custom sort function, or even to pass in a noop.

dahlbyk:
> For what it's worth, my initial instinct was to override the `onSort` event to provide custom sort behavior, but it also seems unavailable at the moment to provide a custom sort function, or even to pass in a noop.

Yeah, `onSort` is composed with `onClick` [here](https://github.com/GriddleGriddle/Griddle/blob/2988f42dd4b3db027587694028783f6507e7848d/src/components/TableHeadingCellEnhancer.js#L13). I also tried a `customHeadingComponent` on the column, but `onClick` is handled by that component's parent.

Events in general feels like its usage is a little unclear, as is its relationship to `actions` (which are imported into `events`).

ryanlanciaux:
Thanks a ton for adding this! Cleans this area up a lot! 

In regard to
> I'm curious if anyone can think of a clean extensibility pattern that would allow either the custom setSortProperties or Heidi's CustomSortTableHeader to access a custom prop on <ColumnDefinition />

 I've been thinking through these kind of scenarios a bit but not really set on a direction so any additional thoughts / pain points etc would be really helpful.

PR #639, opened 2017-04-11
Hover plugin

yoursdearboy:
## Griddle major version
1.3.0

Added a plugin that provide each `Cell` with `hoveredRowKey` property which is equal to `griddleKey` of hovered row.

## Why these changes are made
Don't know if this repo is a proper place to contribute plugins (at least I see plugins directory).
We use it to show toolbar for records in rows (see storybook demo).

## Are there tests?
Storybook story only.


MichaelDeBoey:
@dahlbyk Any news on merging this one?

src/plugins/hover/components.js
```diff
@@ -0,0 +1,43 @@
...
+import { bindActionCreators } from 'redux';
+
+import components from '../../components';
+import { rowMouseEnter, rowMouseLeave } from './actions';
+
+export const RowContainer = compose(
+    connect(
+        state => {
+            return {
+                hoveredRowKey: state.get('hoveredRowKey', null)
```
dahlbyk: I'm not entirely convinced it should be the plugin's responsibility to inject `hoveredRowKey` into `Cell` (or anywhere). It seems to be an implementation detail of this demo. I'm more inclined to avoid customizing `Row` altogether, letting `ButtonsColumn` be connected instead:

```javascript
const ButtonsColumn = connect(
    state => ({ hoveredRowKey: state.get('hoveredRowKey', null) })
  )(
    ({griddleKey,hoveredRowKey}) =>
      (griddleKey === hoveredRowKey) && <button>Edit</button> : null
  );
```

MichaelDeBoey:
@yoursdearboy Any news on addressing @dahlbyk's comments? 🙂

dahlbyk:
@MichaelDeBoey honestly, I would just copy the relevant bits into your own plugin. There's not all that much to wiring up [row events](https://github.com/GriddleGriddle/Griddle/blob/1e6346ab37df96adcd69a22ef2959f7be2f5db00/stories/index.tsx#L169-L177).

MichaelDeBoey:
@dahlbyk Yeah I had a few problems doing so, so was hoping the plugin could help 😕

dahlbyk:
Glad to help as time allows

PR #644, opened 2017-04-19
Implement RowDefinition cssFunction

### LINKED ISSUE #643: Row cssFunction not working
## Griddle version
1.3.1

## Expected Behavior
Row should have class name applied

## Actual Behavior
Class name not applied

## Steps to reproduce
            <RowDefinition
                cssFunction={(rowData) => rowData.active === "A" ? "active" : "inactive"}>
            </RowDefinition>

dahlbyk:
## Griddle major version

1.*

1. `rowDataSelector` is no longer limited to `LocalPlugin`
2. `rowData` and `index` (on current page) have been added to `Row` `props`.
3. `RowDefinition` `cssFunction` is now called if it exists to get a `Row`'s `className`.

The expected signature of `cssFunction` was unclear, so I ultimately opted for accepting the `Row`'s `props`, which seems more flexible than the `rowData` and `index` I initially implemented.

Setting the current implementation aside, I wonder about renaming `cssFunction` as `cssClassName` (aligned with the `ColumnDefinition` prop, and not a breaking change since it has never worked) and allowing both to accept either a function (which we would call with the current `Row`/`Cell` `props`) or a plain string.

## Why these changes are made

Fixes #643

## Are there tests?

No, but there's a story!

### DIFF
--- a/src/components/CellContainer.js
+++ b/src/components/CellContainer.js
@@ -11,6 +11,7 @@ import {
   classNamesForComponentSelector,
   stylesForComponentSelector
 } from '../selectors/dataSelectors';
+import { valueOrResult } from '../utils/valueUtils';
 
 function hasWidthOrStyles(cellProperties) {
   return cellProperties.hasOwnProperty('width') || cellProperties.hasOwnProperty('styles');
@@ -49,7 +50,7 @@ const ComposedCellContainer = OriginalComponent => compose(
   mapProps(props => {
     return ({
     ...props,
-    className: props.cellProperties.cssClassName || props.className,
+    className: valueOrResult(props.cellProperties.cssClassName, props) || props.className,
     style: getCellStyles(props.cellProperties, props.style),
     value: props.customComponent ?
       <props.customComponent {...props} /> :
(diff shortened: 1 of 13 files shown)


dahlbyk:
> ...and allowing both to accept either a function (which we would call with the current Row/Cell props) or a plain string.

Seems there is precedence for this kind of thing from the old [`bodyCssClassName`](https://griddlegriddle.github.io/v0-docs/customization.html#row-metadata).

ryanlanciaux:
> Setting the current implementation aside, I wonder about renaming cssFunction as cssClassName and allowing both to accept either a function or a plain string

I really like that idea.

dahlbyk:
> I really like that idea.

Done.

ryanlanciaux:
Awesome thanks!

wizhippo:
Any change we can get a build pushed for npm?

ryanlanciaux:
Just added griddle-react@1.4.0 to npm. It's not set as release branch so will have to install it griddle-react@1.4.0 👍

PR #653, opened 2017-05-03
Support ColumnDefinition Sortable

### LINKED ISSUE #591: How can I disable sorting for certain columns?

### LINKED ISSUE #652: Default header click handler sorts columns where sortable = false
## Griddle version
1.3.1

## Expected Behavior
If a column has sortable=false, it should not change the sort order when clicking on that column.

## Actual Behavior
It changes the sort order.

https://github.com/GriddleGriddle/Griddle/blob/master/src/plugins/local/components/TableHeadingCellContainer.js#L45

dahlbyk:
## Griddle major version

1.x

`sortable`, which was already defined on `ColumnDefinition`, is now respected.

I also added a default `cursor: pointer` style to `sortable` columns. It occurs to me that there's not a great way to override this that is `sortable`-aware. Should we add `style`/`headerStyle` props to `ColumnDefinition` in imitation of `cssClassName`/`headerCssClassName`?

## Why these changes are made

Fixes #591.
Fixes #652. cc @andreme

## Are there tests?

Story!

### DIFF
--- a/src/components/TableHeadingCellContainer.js
+++ b/src/components/TableHeadingCellContainer.js
@@ -43,11 +43,16 @@ const EnhancedHeadingCell = OriginalComponent => compose(
       <props.customHeadingComponent {...props} icon={icon} /> :
       <DefaultTableHeadingCellContent title={props.title} icon={icon} />;
     const className = valueOrResult(props.cellProperties.headerCssClassName, props) || props.className;
+    const style = {
+      ...(props.cellProperties.sortable === false || { cursor: 'pointer' }),
+      ...props.style,
+    };
 
     return {
       ...props,
       icon,
       title,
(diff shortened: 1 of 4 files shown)


dahlbyk:
So my first iteration was pretty awful. This feels better.

andreme:
@dahlbyk Thanks for fixing this. 

There is comment that can now be removed: https://github.com/GriddleGriddle/Griddle/blob/0b52a4f2e12b6720ca9753d4364b0f8eb05cf340/src/module.d.ts#L56

ryanlanciaux:
This is great! 👍

Issue #651, opened 2017-05-02
Warning: Accessing PropTypes via the main React package


followbl:
@ryanlanciaux I'd be happy to tackle this however you like/recommend...the common approach for now seems to be swapping out for the prop-types package...like so...

https://github.com/gaearon/redux-devtools/pull/358

## Griddle version
1.4.0

## Expected Behavior
console is clean of warnings

## Actual Behavior
warning.js:36 Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.

## Steps to reproduce
Load any of the Griddle Components that include PropTypes (which I believe it all of em)

ryanlanciaux:
That would be amazing if you wanted to take this on! I think that sounds like a good approach

followbl:
will do @ryanlanciaux ...will try and get to it tonight 👍

dahlbyk:
Can this be closed now?

followbl:
Of course. Thanks

skipjack:
Just a heads up on v1.6.0 this warning is thrown again due to the `react-redux-custom-store` package which seems to be a dependency of this package.

moraleslos:
@skipjack +1 

Was just about to post a new issue.  I too have that warning using 1.6.0.

skipjack:
@moraleslos yeah that's a good idea -- not sure if the issue should go here or on the dependency (`react-redux-custom-store`). Maybe both.

PR #654, opened 2017-05-03
fixed PropTypes warnings due to upcoming deprecation in React v15.5

followbl:
https://github.com/GriddleGriddle/Griddle/issues/651

There's still an error in the console on the *storybook*...but after this, it's been determined the culprit is a non-upgraded dependency in @kadira/storybook

```brad@~ Griddle (issue-651) $ yarn why react-inspector
yarn why v0.23.3
[1/4] 🤔  Why do we have the module "react-inspector"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
info This module exists because "@kadira/storybook#@kadira/storybook-addon-actions" depends on it.
info Disk size without dependencies: "208kB"
info Disk size with unique dependencies: "1.23MB"
info Disk size with transitive dependencies: "20MB"
info Amount of shared dependencies: 32
✨  Done in 0.97s.```


followbl:
to be clear, this does fix the console errors in the Griddle package, which was the goal of this PR/issue ;)

followbl:
@dahlbyk yea. funny enough I blew away the yarn.lock and added `prop-types` a second time and it insisted on updating the yarn.lock file like so...I agree, weird

ryanlanciaux:
Awesome thank you!

juampick:
Hi! Thanks for this update. Do you know @ryanlanciaux when are we going to have a new release version? I'm using `1.4.0` and still has this issue. 

Thanks a lot.

PR #658, opened 2017-05-04
Fixed Many-ColumnDefinition Ordering

### LINKED ISSUE #619: columns get somehow shuffled when there are more than 8 columns in ColumnDefinition
## Griddle version

1.1.0 / 1.2.0
chrome & safari (others not tested)

## Expected Behavior

Columns should always be in the same order as defined in `<RowDefinition><ColumnDefinition .. /></RowDefinition>`

## Actual Behavior

if there are more than 8 columns, the order is different. It's always the same, but i did not found the logic how it gets shuffled.

## Steps to reproduce

add more than 8 columnDefinition

## Pull request with failing test or storybook story with issue

I hope i can do this later, but maybe someone has an idea why this is happening. its really odd ;-o

dahlbyk:
## Griddle major version

1.x

It's still not clear to me what precisely is causing the order to change, but it seems wise in general to set column `order` earlier to use everywhere (which fixes the bug).

Also cleaned up `columnUtils` a bit. 

## Why these changes are made

Fixes #619.

## Are there tests?

Story!

### DIFF
--- a/src/utils/columnUtils.js
+++ b/src/utils/columnUtils.js
@@ -1,23 +1,12 @@
-function containsColumnPropertiesArray(rowProperties) {
-  return (rowProperties &&
-    rowProperties.props &&
-    rowProperties.props.children &&
-    Array.isArray(rowProperties.props.children));
-}
-
-function containsColumnPropertiesObject(rowProperties) {
-  return rowProperties && rowProperties.props && rowProperties.props.children;
-}
-
 /** Gets a column properties object from an array of columnNames
  * @param {Array<string>} columns - array of column names
  */
-export function getColumnPropertiesFromColumnArray(columns) {
+export function getColumnPropertiesFromColumnArray(columnProperties, columns) {
   return columns.reduce((previous, current) => {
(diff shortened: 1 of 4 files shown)


src/utils/columnUtils.js
```diff
@@ -1,23 +1,14 @@
...
-
-function containsColumnPropertiesObject(rowProperties) {
-  return rowProperties && rowProperties.props && rowProperties.props.children;
-}
-
 /** Gets a column properties object from an array of columnNames
  * @param {Array<string>} columns - array of column names
  */
-export function getColumnPropertiesFromColumnArray(columns) {
+function getColumnPropertiesFromColumnArray(columnProperties, columns) {
```
dahlbyk: Technically this is a breaking change, but there's really no reason for this to have been used publicly.

ryanlanciaux: Agree that this is more an internal method

ryanlanciaux:
:+1:

andreme:
Thanks @dahlbyk

PR #659, opened 2017-05-05
Suppress implicit any

Jermorin:
## Griddle major version
1.3
Remove implicit any
## Why these changes are made
Because we can't force people to use implicit any mode (`"noImplicitAny": false`) with typescript :smile:

### DIFF
--- a/src/module.d.ts
+++ b/src/module.d.ts
@@ -361,14 +361,14 @@ export namespace utils {
     const rowUtils: PropertyBag<Function>;
 
     interface SortProperties{
-      setSortColumn(sortProperties: ((GriddleSortKey) => void));
+      setSortColumn(sortProperties: ((GriddleSortKey : any) => void)) : any;
       sortProperty: GriddleSortKey;
       columnId: string;
     }
 
     namespace sortUtils {
-      function defaultSort(data: any[], column: string, sortAscending?: boolean);
-      function setSortProperties(sortProperties: SortProperties);
+      function defaultSort(data: any[], column: string, sortAscending?: boolean) : any;
+      function setSortProperties(sortProperties: SortProperties) : any;
     }
 }
 
@@ -400,7 +400,7 @@ export namespace plugins {
     var PositionPlugin : (settings: PositionSettings) => GriddlePlugin;
 }
 
-export const ColumnDefinition;
-export const RowDefinition;
+export const ColumnDefinition : any;
+export const RowDefinition : any;
 
 export default Griddle;
--- a/src/module.d.ts
+++ b/src/module.d.ts
@@ -400,7 +400,7 @@ export namespace plugins {
     var PositionPlugin : (settings: PositionSettings) => GriddlePlugin;
 }
 
-export const ColumnDefinition : any;
-export const RowDefinition : any;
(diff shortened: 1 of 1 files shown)


src/module.d.ts
```diff
@@ -361,14 +361,14 @@ export namespace utils {
...
+      setSortColumn(sortProperties: ((GriddleSortKey : any) => void)) : any;
       sortProperty: GriddleSortKey;
       columnId: string;
     }
 
     namespace sortUtils {
-      function defaultSort(data: any[], column: string, sortAscending?: boolean);
-      function setSortProperties(sortProperties: SortProperties);
+      function defaultSort(data: any[], column: string, sortAscending?: boolean) : any;
+      function setSortProperties(sortProperties: SortProperties) : any;
```
dahlbyk: Can return `() => void`.

src/module.d.ts
```diff
@@ -361,14 +361,14 @@ export namespace utils {
     const rowUtils: PropertyBag<Function>;
 
     interface SortProperties{
-      setSortColumn(sortProperties: ((GriddleSortKey) => void));
+      setSortColumn(sortProperties: ((GriddleSortKey : any) => void)) : any;
```
dahlbyk: `GriddleSortKey` is the correct type; name it `key`? Should return `void`.

src/module.d.ts
```diff
@@ -400,7 +400,7 @@ export namespace plugins {
     var PositionPlugin : (settings: PositionSettings) => GriddlePlugin;
 }
 
-export const ColumnDefinition;
-export const RowDefinition;
+export const ColumnDefinition : any;
+export const RowDefinition : any;
```
Jermorin: Failed to compile.

Error in node_modules/griddle-react/dist/module/module.d.ts
(403,14): error TS7005: Variable 'ColumnDefinition' implicitly has an 'any' type.

Error in node_modules/griddle-react/dist/module/module.d.ts
(404,14): error TS7005: Variable 'RowDefinition' implicitly has an 'any' type.

Jermorin:
> noImplicitAny is probably too arduous to require of our Storybook, but it would be nice if we could fail the build if this type definition regresses. Thoughts?

Yes but i don't know how we can check just the file module.d.ts

src/module.d.ts
dahlbyk: Try `export const ColumnDefinition : ColumnDefinition`?

Jermorin: We can't `export const columnDefinition : ColumnDefinition` ColumnDefinition must be a type.

dahlbyk: Figured out how to make these typed and pushed to your branch.

src/module.d.ts
```diff
@@ -400,7 +400,7 @@ export namespace plugins {
     var PositionPlugin : (settings: PositionSettings) => GriddlePlugin;
 }
 
-export const ColumnDefinition : any;
-export const RowDefinition : any;
+export const ColumnDefinition : typeof components.ColumnDefinition;
+export const RowDefinition : typeof components.RowDefinition;
```
Jermorin: Oh nice !

PR #666, opened 2017-05-24
TypeScript Fix + Pagination with Size Setting

dahlbyk:
## Griddle major version

1.6

1. Cleans up obsolete `Pagination` props
2. Adds an example of moving the page size component into `Pagination`
3. Fixes disabling a settings component by clearing its key on `SettingsComponent`
4. Fixes `GriddleComponent<T>` signature, which had apparently never _really_ been tested
   * Discovered with the next commit...
    * But also discovered by compiling an existing app against Griddle 1.6, which includes https://github.com/GriddleGriddle/Griddle/commit/02be19d29b956a000cc9b2b4cc4cd5583f3891db. `ColumnDefinition` had previously been exported without a type, so the types passed to `customComponent` had not been checked.

## Why these changes are made

Who doesn't love more stories? Also, TypeScript should work.

## Are there tests?

Stories!

### DIFF
--- a/src/components/Pagination.js
+++ b/src/components/Pagination.js
@@ -1,24 +1,14 @@
 import React from 'react';
 
-const Pagination = ({ getPrevious,
-  getNext,
-  setPage,
-  maxPages,
-  currentPage,
-  hasNext,
-  hasPrevious,
-  Next=null,
-  Previous=null,
-  PageDropdown=null,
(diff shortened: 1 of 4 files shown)


ryanlanciaux:
👍 Thanks for catching this. I'm using Griddle in a TypeScript project and kind of sad I didn't notice this :)

Issue #668, opened 2017-06-06
CustomRowComponent does NOT get handed state


sww314:
## Griddle version
1.5 and 1.3 (works in 1.0.3)

## Expected Behavior
Customize row to make it look like a list. The rowDataSelector is not getting handed the state objects (I think).
Follow the code here in the documentation:
https://github.com/GriddleGriddle/griddle-docs/blob/master/components/GriddleList.js

## Actual Behavior
    const CustomRowComponent = connect((state, props) => ({
      rowData: plugins.LocalPlugin.selectors.rowDataSelector(state, props),
    }))(({ rowData }) => (
      <HotelDetail
        hotel={rowData}
        selectHotel={this.selectHotelList}
        favoriteHotel={this.favHotel}
      />
    ));
.....

```
 <Griddle
      data={this.props.hotels}
      pageProperties={{
        pageSize: 60,
      }}
      plugins={[plugins.LocalPlugin]}
      sortProperties={sortProperties}
      components={{
        Row: CustomRowComponent,
        TableContainer: CustomTableComponent,
        TableBody: CustomTableBody,
        SettingsToggle: () => <span />,
      }}
    />
```

TypeError: state.get is not a function
    at Object.rowDataSelector (dataSelectors.js:289)
    at Function.mapToProps (HotelList.js:28)
    at mapToPropsProxy (wrapMapToProps.js:43)
    at Function.detectFactoryAndVerify (wrapMapToProps.js:52)
    at mapToPropsProxy (wrapMapToProps.js:43)
    at handleFirstCall (selectorFactory.js:26)
    at pureFinalPropsSelector (selectorFactory.js:74)
    at Object.runComponentSelector [as run] (connectAdvanced.js:26)
    at Connect.initSelector (connectAdvanced.js:178)
    at new Connect (connectAdvanced.js:119)

## Steps to reproduce
Run the example in the Documentation
https://github.com/GriddleGriddle/griddle-docs/blob/master/components/GriddleList.js
 
## Pull request with failing test or storybook story with issue

While this step is not necessary, a failing test(s) and/or a [storybook story](https://github.com/storybooks/react-storybook) will help us resolve the issue much more easily. Please see the README for more information.

thelittleblacksmith:
We're getting this error as well. For now, removing `plugins.LocalPlugin` from the plugins prop renders the data as a card list but obviously removes the filtering feature.

[...]

PR #676, opened 2017-06-13
Add story with list row demo

dahlbyk:
## Griddle major version

1.x

Adds story from [`GriddleList`](https://github.com/GriddleGriddle/griddle-docs/blob/429f318778604c5e7500c1c949fe1c3137972419/components/GriddleList.js) in the [Griddle docs](http://griddlegriddle.github.io/Griddle/examples/asList/).

## Why these changes are made

Trying to prove https://github.com/GriddleGriddle/Griddle/issues/668

### DIFF
--- a/stories/index.tsx
+++ b/stories/index.tsx
@@ -461,6 +461,61 @@ storiesOf('Griddle main', module)
       }}
     />
   })
+  .add('with list row component', () => {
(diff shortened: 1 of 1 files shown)


ryanlanciaux:
This is a good thing to have a story for and nice to not have to rely on docs to see this in action. Thanks!

Issue #667, opened 2017-05-31
Doc: gallery of themes


cdupont:
With version 1.0.0, unfortunately no styling is applied by default (#661).
The documentation is not so helpful in helping applying quick styles...
Why not making a gallery of themes/quick styles, to help the user pick a styleConfig?

aftabnaveed:
Something I am using at the moment with Bootstrap theme.

```javascript
import React from 'react';
import Griddle, {plugins} from 'griddle-react';

import BootstrapFilter from './BootstrapFilter';

const styleConfig = {
	classNames: {
		Filter: 'form-control input-sm',
		Table: 'table table-bordered table-hover dataTable',
		TableHeading: '',
		TableBody: '',
		PageDropDown: 'form-control input-sm',
		NextButton: 'paginate_button next',
		PreviousButton: 'paginate_button previous',
		Pagination: 'dataTables_paginate paging_simple_numbers'
	}
}

const BootstrapLayout = ({Table, Pagination, Filter, SettingsWrapper}) => {
	return(
		<div className="dataTables_wrapper form-inline dt-bootstrap">
			<div className="row">
				<div className="col-sm-6">
					<div className="dataTables_length">
						<SettingsWrapper />
					</div>
				</div>
				<div className="col-sm-6">
					<div className="dataTables_filter">
						<Filter />
					</div>
				</div>
			</div>
			<div className="row">
				<div className="col-cm-12">
					<Table />
				</div>
			</div>
			<div className="row">
				<div className="col-sm-5">
				</div>
				<div className="col-sm-7">
					<Pagination />
				</div>
			</div>
		</div>
	);
}

class Grid extends React.Component {

	render() {
		return(
			<Griddle 
				{...this.props}
				data={this.props.data} 
				plugins={this.props.plugins} 
				styleConfig={styleConfig} 
				components={{
					Filter: BootstrapFilter,
					Layout: BootstrapLayout
				}}
			/>	
		);
	}
}

export default Grid;
```

dahlbyk:
@aftabnaveed Thanks for sharing this! Can you share the source for `BootstrapFilter` as well?

For theme-creation purposes, it seems like it would be helpful to teach Griddle to merge in `styleConfig` from plugins.

aftabnaveed:
[...]

Issue #661, opened 2017-05-06
Is it supposed to look like this?


jamesskyoung:
## Griddle version
"griddle-react": "^1.0.0",
(Also tried current version)

## Expected Behavior
LIke the site --> https://griddlegriddle.github.io/Griddle/docs/

## Actual Behavior
See screenshot

## Steps to reproduce
Use code from https://griddlegriddle.github.io/Griddle/docs/

Also cloned the current repo... all examples look like there's a missing css etc...

![screen shot 2017-05-06 at 10 04 08](https://cloud.githubusercontent.com/assets/13589738/25771243/dda5474a-3243-11e7-9045-23e89533844b.png)

## Pull request with failing test or storybook story with issue

While this step is not necessary, a failing test(s) and/or a [storybook story](https://github.com/storybooks/react-storybook) will help us resolve the issue much more easily. Please see the README for more information.

jamesskyoung:
Edit: Tried 
```
npm install griddle-react@version0 --save 
```
That works.. (after changing data= to results of course.  

So something's either missing in the dependencies.. or..?

fyi example is this simple.. v1 does not work with this... see screenshot above.  v0.x does. (after simple change)

```
var data = [
    {
    "id": 0,
    "name": "Mayer Leonard",
    "city": "Kapowsin",
    "state": "Hawaii",
    "country": "United Kingdom",
    "company": "Ovolo",
    "favoriteNumber": 7
    }
  
];
    return (
          <Griddle
            results={data}
          />
    )
  }
```

jamesskyoung:
v0 result.
![screen shot 2017-05-06 at 12 37 52](https://cloud.githubusercontent.com/assets/13589738/25772026/e3bb0dbc-3258-11e7-8eea-2d1b1e4091f5.png)

bharadwajag:
Hi ,
 we are also facing same issue for the beta version - "griddle-react": "^1.0.0".

bharadwajag:
@jamesskyoung  
Any idea why it didn't work in the latest version.

jamesskyoung:
None sorry....

jamesskyoung:
I've been using pre V1... but wanted to try the latest version...

sww314:
In version 1+, all the default styling is removed. You have to add your own. 
I use bootstrap styles, so this works for me. 
```
const styleConfig = {
  icons: {
    TableHeadingCell: {
      sortDescendingIcon: '▼',
[...]

PR #677, opened 2017-06-18
Plugin styleConfig and components.Style

dahlbyk:
## Griddle major version

1.x

* `styleConfig` from plugins is now merged in between default styles and the `styleConfig` user prop
* The default `Layout` now renders a `Style` component (default renders `null`) that plugins and users can override to inject a stylesheet

## Why these changes are made

* Gives a mechanism for including a new default "theme" via plugin (#661)
* Gives a mechanism for others to contribute themes (#667)

## Are there tests?

Incomplete story - I pulled in the [CSS from v0](https://github.com/GriddleGriddle/Griddle/blob/359f3f1243b857373b8892a0ff93d60797152f60/docs/assets/styles/griddle.css), but don't know v0 well enough to fix it up to look nice.

### DIFF
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -1,7 +1,8 @@
 import React from 'react';
 
-const component = ({Table, Pagination, Filter, SettingsWrapper, className, style}) => (
+const component = ({Table, Pagination, Filter, SettingsWrapper, Style, className, style}) => (
   <div className={className} style={style}>
+    {Style && <Style />}
     <Filter />
     <SettingsWrapper />
     <Table />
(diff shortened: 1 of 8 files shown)


dahlbyk:
N.B. https://github.com/GriddleGriddle/Griddle/issues/661#issuecomment-299956856 implies the last commit may need to be reverted.

I do think it's important to make Griddle easy to skin, especially when there are common design languages like Bootstrap, Material, Skeleton, etc. That could be one or more plugins in the box, or separate packages.

ryanlanciaux:
I like that it's a plugin. Thanks for adding this -- I totally agree that it would be nice for people to have the ability to skin this easier.

PR #685, opened 2017-07-05
fix of long initialization

klerick:
## Griddle major version 
1.5.0

We got performance issue with `transformData`.
After a half day of researching I fixed the problem and now it runs 50x times faster!
The problem was in `concat` method — it works slow on 10k+ rows.

## Are there tests?
No new tests were added


joellanciaux:
Sorry for the delay and thank you so much for your PR!

It looks like a similar update paired with a handful of other performance changes is included in the latest version (currently `1.8.1`. I'll close this out for now.

Thanks again! :+1:

PR #687, opened 2017-07-07
Typescript Upgrade

### LINKED ISSUE #686: Error while running "npm start" after cloned source-code.
1.5.0

cloning source then npm install, npm run. Recieves errors.

ERROR in ./stories/index.tsx
(197,16): error TS2322: Type '{ data: FakeData[]; plugins: (GriddlePlugin | { events: { setSortProperties: (sortProperties: any...' is not assignable to type 'IntrinsicAttributes & Intrin
sicClassAttributes<Griddle<FakeData>> & Readonly<{ children?: ReactNo...'.
  Type '{ data: FakeData[]; plugins: (GriddlePlugin | { events: { setSortProperties: (sortProperties: any...' is not assignable to type 'Readonly<GriddleProps<FakeData>>'.
    Types of property 'plugins' are incompatible.
      Type '(GriddlePlugin | { events: { setSortProperties: (sortProperties: any) => () => void; }; })[]' is not assignable to type 'GriddlePlugin[]'.
        Type 'GriddlePlugin | { events: { setSortProperties: (sortProperties: any) => () => void; }; }' is not assignable to type 'GriddlePlugin'.
          Type '{ events: { setSortProperties: (sortProperties: any) => () => void; }; }' has no properties in common with type 'GriddlePlugin'.

Why do I get this error, am I missing something? Application is starting fine: 
React Storybook started on => http://localhost:6006/

The error is raised when:
ts-loader: Using typescript@2.4.1 and C:\testprojekt\Griddle\tsconfig.json

dahlbyk:
## Griddle major version

1.x

The good news is TypeScript 2.4 found some type ambiguities.

The bad news is that the React type definitions upgrade isn't backwards-compatible (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/17288), so we either have to upgrade or pin an older version of `@types/react`.

So we upgrade. Fixes #686.

## Are there tests?

The compiler is a test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "devDependencies": {
     "@kadira/storybook": "^2.5.2",
-    "@types/react": "^15.0.17",
+    "@types/react": "^15.0.34",
     "ava": "^0.17.0",
     "babel-cli": "^6.23.0",
     "babel-core": "^6.0.0",
@@ -55,7 +55,7 @@
     "react-dom": "^15.3.2",
     "rimraf": "^2.6.0",
     "ts-loader": "^2.0.2",
-    "typescript": "^2.2.1",
+    "typescript": "^2.4.1",
     "webpack": "^1.14.0",
     "webpack-dev-server": "^1.16.2"
   },
--- a/src/module.d.ts
+++ b/src/module.d.ts
@@ -84,6 +84,8 @@ export class ColumnDefinition extends React.Component<ColumnDefinitionProps, any
 }
 
 export interface CellProps {
+    griddleKey?: number;
+    columnId?: string;
     value?: any;
     onClick?: React.MouseEventHandler<Element>;
     onMouseEnter?: React.MouseEventHandler<Element>;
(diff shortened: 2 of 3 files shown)


ryanlanciaux:
If I understand correctly, we should be on a new major version? @joellanciaux has an inflight change that would require a new major version also. Thoughts?

dahlbyk:
I'm not sure I understand the question. If you're asking if this change would require bumping the Griddle major version, I would say no. This is purely navigating TypeScript + React weirdness; the alternative is for users to update their `package.json` to avoid newer versions of `@types/react`.

ryanlanciaux:
Oh, I think I get it now :)

archy-bold:
When will we get a release of this?

archy-bold:
Also should the types of `griddleKey` on `CellProps` and `RowProps` be different?

dahlbyk:
> Also should the types of `griddleKey` on `CellProps` and `RowProps` be different?

No, they should both be `number`. Good catch!

PR #699, opened 2017-07-22
Sort Icon Classes

### LINKED ISSUE #698: Griddle heading ascending/descending classnames not generated
I'm using Griddle with pagination, filter and local plugin, with _useGriddleStyles = false_ so that I can use custom styles defined externally.  I'm trying to style the ascending/descending icons of the header to those from _fontawesome_.  I'm doing this by overriding the _griddle-heading-ascending_ and _griddle-heading-descending_ classes like this:

```
.griddle-heading-ascending:before {
   content: "\f15d";
   font-family: FontAwesome;
}
```

However griddle just uses the built-in up/down arrow icons.  Looking at the elements via Chrome, it seems that there are no _griddle-heading-ascending_ and _griddle-heading-descending_ classes generated.  It just creates an inline `<span>` inside the` <th>` with the _griddle-table-heading-cell_ classname:

screenshot_3

Is there supposed to be _griddle-heading-ascending_ and _griddle-heading-descending_ span classes generated?

BTW using Griddle 1.6.0

dahlbyk:
## Griddle major version

1.x

Applies `TableHeadingCellAscending` and `TableHeadingCellDescending` as `className` on the `<span>` that surrounds the sort icon.

Also, extracts a `getSortIconProps` utility function that can be shared between the default and local plugin `TableHeadingCellContainer` implementations.

## Why these changes are made

Fixes #698
cc @moraleslos

## Are there tests?

Story!

### DIFF
--- a/src/components/TableHeadingCellContainer.js
+++ b/src/components/TableHeadingCellContainer.js
@@ -6,6 +6,7 @@ import mapProps from 'recompose/mapProps';
 import getContext from 'recompose/getContext';
 
 import { sortPropertyByIdSelector, iconsForComponentSelector, classNamesForComponentSelector, stylesForComponentSelector, customHeadingComponentSelector, cellPropertiesSelector } from '../selectors/dataSelectors';
+import { getSortIconProps } from '../utils/sortUtils';
 import { valueOrResult } from '../utils/valueUtils';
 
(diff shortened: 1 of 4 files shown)


ryanlanciaux:
Thanks for adding this!

PR #715, opened 2017-08-10
Add "griddleConnect" to help avoid store conflicts

ryanlanciaux:
## Griddle major version
1

This adds a griddle specific connect and storeName prop that all of the container components in Griddle can use (it will be on the module as well when this PR is more finalized).

## Why these changes are made
The thinking here is that it would be nice to use Griddle inside of apps that also are using redux and not have store conflicts with custom components, etc.

## Are there tests?
story for now

### DIFF
--- a/src/components/griddleConnect.js
+++ b/src/components/griddleConnect.js
@@ -1,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux-custom-store';
+import getContext from 'recompose/getContext';
+
+
+const griddleConnect = (...connectOptions) => OriginalComponent => class extends React.Component {
+  static contextTypes = {
+    storeName: PropTypes.string,
+  }
+
+  render() {
+    console.log(this.context.storeName)
+    const ConnectedComponent = connect(...connectOptions)(OriginalComponent, this.context.storeName);
+
+    return <ConnectedComponent {...this.props}/>
+  }
+}
+
+export { griddleConnect as connect }
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "lodash": "^4.17.4",
     "max-safe-integer": "^1.0.0",
     "prop-types": "^15.5.8",
-    "react-redux": "^5.0.2",
+    "react-redux": "^5.0.6",
     "react-redux-custom-store": "^1.0.0",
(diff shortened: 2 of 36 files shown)


StevenTCramer:
This will be a big help for us.  If we can get this through.  

Thanks!  Pizza Pizza.

ryanlanciaux:
I think this is okay to keep as 1.x.0 instead of 2.0.0 -- per discussion you have to specify a new store name to change the default behavior from what it was.

Any thoughts on all this would be greatly appreciated.

centerwork:
I really could use this too.

src/components/griddleConnect.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux-custom-store';
+import getContext from 'recompose/getContext';
```
ryanlanciaux: Ahh yea it is unused :D

ryanlanciaux:
Totally agree about the placement of it under components being weird -- I'll move this around. I will have to look more into the patch for createProvider and see if it's something that could be done simply without that 3rd party library

sww314:
Not sure how it is done under the covers but redux-form has a good dev experience for connection to the Redux store.

dahlbyk:
> I will have to look more into the patch for createProvider and see if it's something that could be done simply without that 3rd party library

This is half of the picture. The other half, for which I believe `react-redux-custom-store` is still needed, is the replacement for `connect` that uses the correct store. `react-redux` does export [`createConnect`](https://github.com/reactjs/react-redux/blob/fd81f1812c2420aa72805b61f1d06754cb5bfb43/src/connect/connect.js#L40) to achieve modified `connect` behavior, but I'm not certain they expose quite what we need. Maybe wrapping [`connectAdvanced`](https://github.com/reactjs/react-redux/blob/fd81f1812c2420aa72805b61f1d06754cb5bfb43/src/components/connectAdvanced.js) to inject `storeKey` and passing that as `connectHOC` in `createConnect`?

ryanlanciaux:
Looks like we can use connect and pass in `storeKey` without any third-party module.

ryanlanciaux:
I _think_ this one should be ready. Once it is g2g, I'll work on getting a npm release going.

PR #717, opened 2017-08-13
onSort event argument

### LINKED ISSUE #716: onSort event not giving sortAscending ascending property
## Expected Behavior
We can also pass custom events in order to controlling paging, sorting etc...
I am facing issue with onSort event, It will return sortProperty with **id** and **sortAscending** property. like
If i am clicking on **name** column then onSort should return:
`{ id: 'name', sortAscending: true }`
and if i again click on **name** column then onSort should return:
`{ id: 'name', sortAscending: false }`

## Actual Behavior
but actually it is returning only **id** property in the both case not the **sortAscending**
`{ id: 'name' }`

## Steps to reproduce
1. Griddle with static data
2. Pass events to griddle

```
 let pagingEvents = {
     onFilter: (filterText) = > {
         console.log("filterText", filterText)
     },
     onSort: (sortProperties) = > {
         console.log("sortProperties", sortProperties)
     },
     onNext: () = > {
         console.log("on Next callback")
     },
     onPrevious: () = > {
         console.log("on Prev callback")
     },
     onGetPage: (pageNumber) = > {
         console.log("pageNumber", pageNumber)
     },
 }

<Griddle data={results} plugins={[plugins.LocalPlugin]} styleConfig={styleConfig} events={pagingEvents}>
    <RowDefinition>
        <ColumnDefinition id="name" title="Name" width={200} style={{ textAlign: 'center' }} />
        <ColumnDefinition id="gender" title="Gender" width={80} style={{ textAlign: 'center' }} />
        <ColumnDefinition id="height" title="Height" width={50} style={{ textAlign: 'center' }} />
    </RowDefinition>
</Griddle>

```

dahlbyk:
## Griddle major version

1.x

Pass full `sortProperty` to `events.onSort`, if available.

Note that the initial call only includes `id`, not `sortAscending`. Easy enough to hard-code that to `true`, as in [`setSortProperties`](https://github.com/GriddleGriddle/Griddle/blob/1d7b763215c016fa73d700deec98d5edc56fcbe7/src/utils/sortUtils.js#L30), but it _may_ be useful to be able to identify if this was the initial sort?

## Why these changes are made

It is right and good, and also closes #716 (cc @jariwalabhavesh)

## Are there tests?

`/shrug`

### DIFF
--- a/src/components/TableHeadingCellEnhancer.js
+++ b/src/components/TableHeadingCellEnhancer.js
@@ -11,7 +11,10 @@ const EnhancedHeadingCell = OriginalComponent => compose(
   }),
   mapProps(({ events: { onSort }, ...props }) => ({
     ...props,
-    onClick: combineHandlers([() => onSort && onSort({ id: props.columnId }), props.onClick]),
+    onClick: combineHandlers([
+      () => onSort && onSort(props.sortProperty || { id: props.columnId }),
+      props.onClick,
+    ]),
   }))
 )(props => <OriginalComponent {...props} />);
 
--- a/stories/index.tsx
+++ b/stories/index.tsx
@@ -90,11 +90,11 @@ storiesOf('Griddle main', module)
 
     // don't do things this way - fine for example storybook
     const events = {
-      onFilter: () => console.log('onFilter'),
-      onSort: () => console.log('onSort'),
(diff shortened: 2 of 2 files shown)


ryanlanciaux:
:+1: Thanks for cleaning that up!

Issue #466, opened 2016-08-23
Sorting is case sensitive


matias-piin:
Hi,

I'm having an issue with the column sorting feature. It seems to be case sensitive, without an option to make it case insensitive.

Example:
![image](https://cloud.githubusercontent.com/assets/15108415/17903297/03c245d8-6939-11e6-89bc-543fdb8b06fd.png)

_I've read through the docs and other issues and could not find a mention of this. Please let me know if this has already been addressed and I will close the issue._

ochervak:
+1

akshayvenugopal:
+1

dahlbyk:
@ryanlanciaux @joellanciaux sort is case-sensitive again with `LocalPlugin` in 1.x. Overriding with `sortMethod` works, but needing this on every column is less than ideal. Glad to submit a patch, but curious if you have any thoughts on where a global default `sortMethod` should live?

dahlbyk:
> @ryanlanciaux @joellanciaux sort is case-sensitive again with LocalPlugin in 1.x. Overriding with sortMethod works, but needing this on every column is less than ideal. Glad to submit a patch, but curious if you have any thoughts on where a global default sortMethod should live?

Bump, before I just spike something out!

vineela-m:
Any updates onto this ?

augiaugi:
@dahlbyk 

Hello, is there any update on this? 
I am missing the commit in v1.8 to use .ToLower() for the sorting.

https://github.com/GriddleGriddle/Griddle/pull/516/commits/34cea1e984c7b2707ea6386a865ee6757988df99

dahlbyk:
> @augiaugi: Hello, is there any update on this?
> I am missing the commit in v1.8 to use .ToLower() for the sorting.
> 
> [34cea1e](https://github.com/GriddleGriddle/Griddle/commit/34cea1e984c7b2707ea6386a865ee6757988df99)

`ToLower()` is only in v0; as of 1.8 (#718) you can specify sort behavior for the whole grid with `<Griddle sortMethod={fn}>`, or per-column sorting with `<ColumnDefinition sortMethod={fn} />`. An example sort function:

https://github.com/NewBoCo/Griddle/blob/7ecddf1ec5ab8d933bb2fbc544788c251f691829/stories/index.tsx#L27-L42

PR #718, opened 2017-08-14
Default sortMethod (plus: initialState from extra props)

dahlbyk:
## Griddle major version

1.x

- Type cleanup for `RowRenderProperties` and `ColumnRenderProperties`, which currently the same as `RowDefinitionProps` and `ColumnDefinitionProps`.
- Rearranges how `initialState` is initialized to accept/override `initialState` from extra `Griddle` props. Everything still seems to work as expected, but please double-check this.
   * As a bonus, this allows setting `enableSettings` and `textProperties` without a plugin, as demonstrated with new stories.
- Default `sortMethod` from state is used, if it exists.
- Addressed TODO to get Next/Previous label from `textProperties`

## Why these changes are made

`<Griddle sortMethod={...}>` now works as a high-level hook to override all columns' sort behavior. It does not change default behavior (cc #466).

## Are there tests?

Stories!

### DIFF
--- a/stories/index.tsx
+++ b/stories/index.tsx
@@ -1214,6 +1214,20 @@ storiesOf('Settings', module)
       <Settings settingsComponents={components} />
     );
   })
+  .add('disable settings', () => {
+    return (
+      <Griddle data={fakeData} plugins={[LocalPlugin]}
+        enableSettings={false}
+        />
+    );
+  })
+  .add('change settings toggle button text', () => {
+    return (
+      <Griddle data={fakeData} plugins={[LocalPlugin]}
+        textProperties={{ settingsToggle: 'Toggle!' }}
+        />
+    );
+  })
   .add('remove built-in settings', () => {
     const plugin = {
       components: {
--- a/stories/index.tsx
+++ b/stories/index.tsx
@@ -1381,6 +1381,12 @@ storiesOf('Settings', module)
           </div>
         ),
       },
+      initialState: {
+        textProperties: {
+          next: '▶',
+          previous: '◀',
+        },
+      },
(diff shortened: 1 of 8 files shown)


ryanlanciaux:
Ahh shoot -- didn't mean to make conflicts here.

dahlbyk:
> Ahh shoot -- didn't mean to make conflicts here.

It's fine, I knew it was coming.

dahlbyk:
(Conflicts have been resolved.)

stories/index.tsx
```diff
@@ -178,6 +178,19 @@ storiesOf('Griddle main', module)
     )
   })
 
+  .add('with custom default sort', () => {
+    return (
```
ryanlanciaux: Should this have the sortProperties set like the previous story?

```
const sortProperties = [
  { id: 'name', sortAscending: true }
];
...
<Griddle
  ...
  sortProperties={sortProperties}
  sortMethod={sortBySecondCharacter}>
```

ryanlanciaux:
This looks great! I've run this locally on some of my projects also and everything seems to work well. I just have the one question on the story but it looks good to merge.

stories/index.tsx
dahlbyk: I don't think this grid needs to be pre-sorted, no. This test is derived from the [test that follows](https://github.com/NewBoCo/Griddle/blob/7ecddf1ec5ab8d933bb2fbc544788c251f691829/stories/index.tsx#L194-L206).

ryanlanciaux: Got it - sorry about my confusion. 👍

PR #721, opened 2017-08-15
Add connect to utils out

ryanlanciaux:
## Griddle major version
1

Add the Griddle connect to something that can be used outside of Griddle.

## Why these changes are made
When making custom components, it's useful to still be able to easily connect to Griddle's store.

## Are there tests?
Not really but the previous PR utilizes connect directly and I am running one of my projects against a local copy of Griddle with this change (using Griddle's connect)


dahlbyk:
This should be added to the type definition, too.

ryanlanciaux:
Whoops -- adding now

PR #723, opened 2017-08-17
Pass down TableBody style and className

JesseFarebro:
## Griddle major version
1.5.0

className and style are fetched from the store but never passed down to the original component.

## Why these changes are made
If a user wants to keep the original TableBody container but wants to override the style or className.

## Are there tests?
No


dahlbyk:
Thanks!

PR #724, opened 2017-08-22
Add Storybook to build + fix connect type definition

dahlbyk:
## Griddle major version

1.8

* [x] Make build fail if TypeScript Storybook won't build.
    * Not sure if there's really any value to targeting `dist/examples/storybook.js` rather than throwing the output away.
    * It seems like `build-storybook` should have been sufficient, but it doesn't recognize JSX. So we're going with a straight webpack for now.
* [ ] Fix type definition for new `connect` - I didn't have a chance to test #722, but it's not happy

## Why these changes are made

1. The build should break if types are broken.
2. The types should not be broken.

## Are there tests?

Kind of?

### DIFF
--- a/src/module.d.ts
+++ b/src/module.d.ts
@@ -419,15 +419,7 @@ export namespace utils {
     const dataUtils: PropertyBag<Function>;
     const rowUtils: PropertyBag<Function>;
 
-    // TODO: Not a fan of this but was trying to avoid 
-    // wholesale copying the 11 or so types out of definitely-typed's connect
-    // not immediately seeing how to say for this thing, just reference their types
-    function connect(
-        mapStateToProps?: ReactRedux.MapStateToPropsParam<any, any>,
-        mapDispatchToProps?: ReactRedux.MapDispatchToPropsParam<any, any>,
-        mergeProps?: ReactRedux.MergeProps<any, any, any, any>,
-        options?: ReactRedux.Options
-    ) : ReactRedux.InferableComponentEnhancerWithProps<any, any>;
+    export const connect : typeof ReactRedux.connect;
 
     interface SortProperties{
       setSortColumn(sortProperties: ((key : GriddleSortKey) => void)) : void;
--- a/src/module.d.ts
+++ b/src/module.d.ts
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import * as ReactRedux from 'react-redux';
+import { connect as originalConnect } from 'react-redux';
 import { ActionCreator } from 'redux';
 
 interface PropertyBag<T> {
@@ -420,7 +420,7 @@ export namespace utils {
     const dataUtils: PropertyBag<Function>;
     const rowUtils: PropertyBag<Function>;
 
-    export const connect : typeof ReactRedux.connect;
+    const connect : typeof originalConnect;
 
     interface SortProperties{
       setSortColumn(sortProperties: ((key : GriddleSortKey) => void)) : void;
--- a/src/module.d.ts
+++ b/src/module.d.ts
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import * as ReactRedux from 'react-redux';
+import { ActionCreator } from 'redux';
 
 interface PropertyBag<T> {
     [propName: string]: T;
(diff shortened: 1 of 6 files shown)


stories/index.tsx
```diff
@@ -7,12 +7,13 @@ import getContext from 'recompose/getContext';
...
 import withHandlers from 'recompose/withHandlers';
 import withState from 'recompose/withState';
-import { Provider, connect } from 'react-redux';
+import { Provider } from 'react-redux';
 import { createStore } from 'redux';
 import { createSelector } from 'reselect';
 import _ from 'lodash';
 
 import GenericGriddle, { actions, components, selectors, plugins, utils, ColumnDefinition, RowDefinition } from '../src/module';
+const { connect } = utils;
```
dahlbyk: I'm unsure if this change is useful/necessary.

src/module.d.ts
```diff
@@ -293,7 +294,7 @@ export interface GriddleComponents {
     PreviousButtonContainerEnhancer?: (OriginalComponent: GriddleComponent<any>) => GriddleComponent<any>;
 }
 
-export interface GriddleActions extends PropertyBag<Function> {
+export interface GriddleActions extends PropertyBag<ActionCreator<any>> {
```
dahlbyk: Apparently this was the root cause of the Storybook build error.

src/module.d.ts
```diff
@@ -419,15 +420,7 @@ export namespace utils {
...
-    // TODO: Not a fan of this but was trying to avoid 
-    // wholesale copying the 11 or so types out of definitely-typed's connect
-    // not immediately seeing how to say for this thing, just reference their types
-    function connect(
-        mapStateToProps?: ReactRedux.MapStateToPropsParam<any, any>,
-        mapDispatchToProps?: ReactRedux.MapDispatchToPropsParam<any, any>,
-        mergeProps?: ReactRedux.MergeProps<any, any, any, any>,
-        options?: ReactRedux.Options
-    ) : ReactRedux.InferableComponentEnhancerWithProps<any, any>;
+    export const connect : typeof ReactRedux.connect;
```
ryanlanciaux: Didn't realize you could do this! Thanks!

stories/index.tsx
ryanlanciaux: I don't think it's necessary but may be good to use Griddle's connect here like this change introduces. That way if there is something that is going astray there it may be a little more clear. :+1:

ryanlanciaux:
I will try to upgrade in one of my projects that uses TypeScript today :| I did from an `npm pack`ed version and on npm before making 1.8.0 the `release` version.

src/module.d.ts
dahlbyk: Full disclosure: I didn't either until I tried it. 😀

dahlbyk: I don't think this is quite right... now getting on a real project:

> Namespace '"./node_modules/@types/react-redux/index"' has no exported member 'connect'.

src/module.d.ts
```diff
@@ -419,15 +420,7 @@ export namespace utils {
...
-    // TODO: Not a fan of this but was trying to avoid 
-    // wholesale copying the 11 or so types out of definitely-typed's connect
-    // not immediately seeing how to say for this thing, just reference their types
-    function connect(
-        mapStateToProps?: ReactRedux.MapStateToPropsParam<any, any>,
-        mapDispatchToProps?: ReactRedux.MapDispatchToPropsParam<any, any>,
-        mergeProps?: ReactRedux.MergeProps<any, any, any, any>,
-        options?: ReactRedux.Options
-    ) : ReactRedux.InferableComponentEnhancerWithProps<any, any>;
+    const connect : typeof originalConnect;
```
dahlbyk: This adjustment seems to have fixed the project I'm upgrading, but I'm unclear why this isn't equivalent to `typeof ReactRedux.connect`.

dahlbyk:
@ryanlanciaux following up on this... 1.8.0 is essentially broken with TS without these fixes. There are a few other bigger changes pending, but could you ship a 1.8.1 release if this looks good?

ryanlanciaux:
Absolutely - this looks good to me. Merging and will ship a new version tonight.

ryanlanciaux:
Hmm may be tomorrow before shipping new version -- I'm currently getting this https://travis-ci.org/GriddleGriddle/Griddle/builds/270145494 because I didn't test against this ahead of latest master, I think (sorry!).

I'll fix the part that's causing the failure and then push a new version.

dahlbyk:
Fixed: https://github.com/GriddleGriddle/Griddle/pull/732

ryanlanciaux:
1.8.1 is out but not pushed as latest -- I'm testing on some projects before doing that.

There _might_ be an issue with the new griddle-connect that I'm trying to track down as well BUT if not, I'll push this one as latest very soon (won't go a long time like pre 1.8 again where there were many releases but none promoted).

PR #726, opened 2017-08-22
Add redux middleware to Griddle props to be passed to the store upon …

shorja:
…creation.

## Griddle major version
1.8.0

Adding a prop to top level Griddle component which passes in user supplied redux middleware which is then applied during redux store creation using applyMiddleware

## Why these changes are made
Running into cases where it would be really great to be able to apply redux middleware and there isn't an easy way to do that right now

## Are there tests?
No

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,4 @@
-import { createStore, combineReducers, bindActionCreators } from 'redux';
+import { createStore, combineReducers, bindActionCreators, applyMiddleware } from 'redux';
 import Immutable from 'immutable';
 import { createProvider } from 'react-redux';
 import React, { Component } from 'react';
@@ -78,6 +78,7 @@ class Griddle extends Component {
       renderProperties:userRenderProperties={},
       settingsComponentObjects:userSettingsComponentObjects,
       storeKey = 'store',
+      reduxMiddleware = [],
       ...userInitialState
     } = props;
 
@@ -134,7 +135,8 @@ class Griddle extends Component {
 
     this.store = createStore(
       reducers,
-      initialState
+      initialState,
+      applyMiddleware(...reduxMiddleware)
     );
 
     this.provider = createProvider(storeKey);
--- a/src/index.js
+++ b/src/index.js
@@ -136,7 +136,7 @@ class Griddle extends Component {
     this.store = createStore(
       reducers,
       initialState,
-      applyMiddleware(...reduxMiddleware)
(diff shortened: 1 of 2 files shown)


dahlbyk:
It would be great if you could include `reduxMiddleware` from plugins as well.

Bonus points for updating the TypeScript definition, too.

JesseFarebro:
@dahlbyk updated with the changes, anything else?

src/module.d.ts
```diff
@@ -396,6 +396,7 @@ export interface GriddlePlugin extends GriddleExtensibility {
 
 export interface GriddleProps<T> extends GriddlePlugin, GriddleInitialState {
     plugins?: GriddlePlugin[];
+    reduxMiddleware?: ReactRedux.Middleware[];
```
dahlbyk: You can move this to `GriddlePlugin`, since it is available for both.

src/index.js
```diff
@@ -134,7 +135,8 @@ class Griddle extends Component {
 
     this.store = createStore(
       reducers,
-      initialState
+      initialState,
+      applyMiddleware(...plugins.map(p => p.reduxMiddleware), ...reduxMiddleware) 
     );
```
dahlbyk: I don't remember… does spreading an array also flatten nested arrays? If it doesn't, this will give us arrays of middleware as arguments. Does that work?

src/module.d.ts
JesseFarebro: Will do! 👍

src/index.js
JesseFarebro: That's very true, I'll flatten them as well!

I compacted it as well (removing undefined from the mapping) because `applyMiddleware` will try to call even undefined functions. See the latest file changes.

dahlbyk:
> Also would you be opposed to composing `applyMiddleware` with redux-devtools if were not bundling for production? It would be really helpful and a quick change.

I don't have any problem with this, but I'll defer to @ryanlanciaux. Maybe submit a follow-up PR for that once this is merged?

src/index.js
dahlbyk: (To clarify, we need both `compact` and `flatten`, correct?)

JesseFarebro: Yes, correct. Flatten the plugins `reduxMiddleware` then remove and undefined values from that (for example if a plugin doesn't define `reduxMiddleware`).

ryanlanciaux:
This looks great! Thanks for adding this.

PR #729, opened 2017-08-23
Improve performance

andreme:
## Griddle major version
1.8

## Why these changes are made
To make loading rows faster

## Are there tests?
Yes

### DIFF
--- a/src/reducers/dataReducer.js
+++ b/src/reducers/dataReducer.js
@@ -1,5 +1,4 @@
 import Immutable from 'immutable';
-import _ from 'lodash';
 
 /*
  * State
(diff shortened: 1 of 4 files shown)


src/reducers/dataReducer.js
```diff
@@ -134,12 +133,10 @@ export function GRIDDLE_TOGGLE_COLUMN(state, action) {
 }
 
 export function GRIDDLE_UPDATE_STATE(state, action) {
-  const transformedData = transformData(action.newState.data, state.get('renderProperties').toJSON());
-  const data = transformedData.data;
-  const lookup = transformedData.lookup;
-  const newState = _.omit(action.newState, data);
```
andreme: _.omit() accepts an array and then tries to unset the each value in the array from the object

src/utils/dataUtils.js
```diff
@@ -34,7 +34,10 @@ export function transformData(data, renderProperties) {
     throw new Error(`Griddle: Property '${renderProperties.rowProperties.rowKey}' doesn't exist in row data. Please specify a rowKey that exists in <RowDefinition>`);
   }
 
-  const transformedData = data.reduce(({ list, lookup }, rowData, index) => {
+  const list = [];
+  const lookup = {};
+
+  data.forEach((rowData, index) => {
```
andreme: this reduces the time spent here for 400 records from 30ms to < 0.3ms

ryanlanciaux: Awesome!

src/reducers/dataReducer.js
dahlbyk: Great use of spread to refactor this!

ryanlanciaux:
Thanks so much for adding this!

PR #730, opened 2017-08-24
Update ColumnChooser

ahopkins:
## Griddle major version

v1

Add `id` attribute to the `input` 

## Why these changes are made

To allow for clicking of label to toggle the checkbox

## Are there tests?

No, it is a single HTML attribute

### DIFF
--- a/src/settingsComponentObjects/ColumnChooser.js
+++ b/src/settingsComponentObjects/ColumnChooser.js
@@ -57,6 +57,7 @@ return (
           <input
             type="checkbox"
             name={hiddenColumns[v].id}
+            id={hiddenColumns[v].id}
             onChange={onToggle}
             defaultChecked={false}
           />


dahlbyk:
> To allow for clicking of label to toggle the checkbox

Are you sure this is necessary? `<label><input /> Test</label>` should already work; `id`+`for` is only necessary if the `input` is outside the `label` ([W3C](https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-implicitly)).

ahopkins:
It was not working for me at all in Firefox Dev Version. I'll send you the
specifics of the browser.

dahlbyk:
> It was not working for me at all in Firefox Dev Version. I'll send you the
> specifics of the browser.

No need; I've confirmed what you're seeing. I think the better fix would be to remove `htmlFor` from both `label`s.

PR #732, opened 2017-08-31
Fix Middleware in type definition

dahlbyk:
## Griddle major version

1.8.0

Fixes type definition bug in #726, now (fortunately) checked by #727.

## Why these changes are made

We like green builds and working type definitions.

## Are there tests?

Types!


ryanlanciaux:
You beat me to it :D 👍

PR #735, opened 2017-09-07
Store listeners proposal

shorja:
## Griddle major version
1.8.0

Adding the ability to have listeners on the store independent of the "visual" part of Griddle. I am envisioning these to be used for running callbacks from the outside world, for example those provided on the "events" object.

## Why these changes are made
Instead of having an event callback passed down to the component that would trigger it, for example on the pagination buttons, a listener function can subscribe to the store and trigger the "onNext" event when the pagination state has changed in the store. This would make for a better separation of concerns between how the state is changed and what has changed. 

I am running into use cases where I want to be listening on the change of a particular plugin's store subset to run an externally provided call-back but would have to hook into UI events in several places to make this work. It would be great if there was just a single place to keep this sort of behaviour.

And of course it's not required to make all events use this design, this is just supplementary.

This is also not a ready-to-merge PR, I have a question: right now I am only getting listeners from plugins, do we want to allow listeners to be passed in as props on <Griddle /> as well? And in that case, will have to update typescript stuff too.
 
## Are there tests?
No

### DIFF
--- a/src/utils/listenerUtils.js
+++ b/src/utils/listenerUtils.js
@@ -1,0 +1,46 @@
+export const StoreListener = class StoreListener {
+    constructor(store) {
+        this.store = store;
+        this.unsubscribers = {};
+    }
+
+    removeListener = (name) => {
+        if (this.unsubscribers.hasOwnProperty(name)) {
+            this.unsubscribers[name]();
+            delete this.unsubscribers[name];
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    // Adds a listener to the store.
+    // Will attempt to remove an existing listener if the name
+    // matches that of an existing listener.
+    // If no name is provided this is an anonymous lister, it
+    // is not registered in the list of unsubscribe functions,
+    // returns the unsubscribe function so it can still be handled
+    // manually if desired.
+    addListener = (listener, name, otherArgs) => {
+        // attempt to unsubscribe an existing listener if the new 
+        // listener name matches
+        // if no name is provided, do nothing
+        name && this.removeListener(name);
+        const unsubscribe = (() => {
+            let oldState;
+            return this.store.subscribe(() => {
+                const newState = this.store.getState();
+                listener(oldState, newState, {...otherArgs});
+                oldState = newState;
+            });
+        })();
+        // if name was provided, add the unsubscribe
+        // otherwise this is an "anonymous" listener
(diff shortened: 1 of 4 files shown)


ryanlanciaux:
I bumped into something last night where this functionality would be really useful. I like this -- I personally would prefer being able to add them as props as well (and would require typescript updates for the .d.ts stuff)

It would be really cool if you had the time to make a story in storybook for wiring this up.

shorja:
Yeah, I'll add them to the props as well.

I was also thinking it would be handy to pass down the list of unsubscribe functions returned by store.subscribe on the context so that one could dynamically unsubscribe their listeners if needed. And move the subscribe function out into the utils so that they can be added as well.

Since the unsubscribes functions would need to be referenced somehow, for example in an object, this could cause name collisions if more than one listener was given the same name. I don't know if thats worth solving for right now?

shorja:
- Added a listeners prop to Griddle which takes an object of listener functions.
- There is now a storeListener on the context which can be used to add and remove listeners dynamically in the components
- Added typescript stuff, a review of this would be great
- Added a storybook example, do we want more than one?

ryanlanciaux:
Thanks a ton for adding this! Will take a look - I think one storybook example is good 😄 

I'll take a look at the TypeScript stuff but I've only been lightly using it for a couple months so if others want to weigh in too that'd be awesome.

src/utils/listenerUtils.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    // Adds a listener to the store.
+    // Will attempt to remove an existing listener if the name
+    // matches that of an existing listener.
+    // If no name is provided this is an anonymous lister, it
+    // is not registered in the list of unsubscribe functions,
+    // returns the unsubscribe function so it can still be handled
+    // manually if desired.
+    addListener = (listener, name, otherArgs) => {
```
ryanlanciaux: This is very cool

ryanlanciaux:
I'm going to test on a couple more projects tomorrow and merge if all is good :+1:

ryanlanciaux:
👍  This is an awesome feature

ryanlanciaux:
This is in Griddle@1.9.0 - which I've tested on a handful of my larger projects (and storybook) and all seems good. Thanks again

1.9.0 is now latest on npm

PR #740, opened 2017-09-18
Change all containers in base components to use context for selectors

shorja:
## Griddle major version
1.9.0

Convert all references to dataSelectors file in components to the context.selectors

## Why these changes are made
This allows plugins to override selectors in the base components without having to wholesale copy them. Would this potentially be a dangerous change from a consumers of Griddle perspective?

## Are there tests?
No

### DIFF
--- a/src/components/TableBodyContainer.js
+++ b/src/components/TableBodyContainer.js
@@ -5,20 +5,22 @@ import compose from 'recompose/compose';
 import mapProps from 'recompose/mapProps';
 import getContext from 'recompose/getContext';
 
-import { visibleRowIdsSelector, classNamesForComponentSelector, stylesForComponentSelector } from '../selectors/dataSelectors';
+//import { visibleRowIdsSelector, classNamesForComponentSelector, stylesForComponentSelector } from '../selectors/dataSelectors';
 
 const ComposedTableBodyContainer = OriginalComponent => compose(
   getContext({
     components: PropTypes.object,
     selectors: PropTypes.object,
   }),
-  connect((state, props) => ({
-    visibleRowIds: visibleRowIdsSelector(state),
-    className: classNamesForComponentSelector(state, 'TableBody'),
-    style: stylesForComponentSelector(state, 'TableBody'),
-  })),
+  connect(
+    (state, props) => ({
+      visibleRowIds: props.selectors.visibleRowIdsSelector(state),
+      className: props.selectors.classNamesForComponentSelector(state, 'TableBody'),
+      style: props.selectors.stylesForComponentSelector(state, 'TableBody'),
+    })
+  ),
   mapProps(props => {
-    const { components, ...otherProps } = props;
+    const { components, selectors, ...otherProps } = props;
     return {
       Row:  props.components.Row,
       ...otherProps,
@@ -26,10 +28,10 @@ const ComposedTableBodyContainer = OriginalComponent => compose(
   }),
 )(({Row, visibleRowIds, style, className}) => (
   <OriginalComponent
-    rowIds={visibleRowIds}
-    Row={Row}
-    style={style}
-    className={className}
+  rowIds={visibleRowIds}
+  Row={Row}
+  style={style}
+  className={className}
   />
 ));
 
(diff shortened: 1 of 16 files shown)


dahlbyk:
Love it! I think we can pretty confident this isn't a breaking change, simply because we're not changing the exported selectors that plugins have had to reference until now. This does make selectors an even more explicit part of the public API, though, so we'll need to be more careful about 1) deciding which to export, and 2) how they are designed (esp. with regard to exposing the immutable internals).

While we're at it, any thoughts on avoiding [this pattern](https://github.com/GriddleGriddle/Griddle/blob/4bdc65e4c8b377751e9e2104ee4b53efa88e54f3/src/plugins/local/selectors/localSelectors.js#L186-L196)?

joellanciaux:
Very cool :+1:

@dahlbyk regarding the extra `dataSelectors` exports from the `localSelectors`, we _should_ be able to simply remove them for the next version.  Would that be worthwhile to include with this update?

dahlbyk:
On further consideration, that particular pattern isn't the problem (though it will be nice to get rid of all the copy/paste). I'm glad to contribute that cleanup if @shorja doesn't wish to add it to this PR.

The pattern I meant to ask about manifests [here](https://github.com/GriddleGriddle/Griddle/blob/4bdc65e4c8b377751e9e2104ee4b53efa88e54f3/src/plugins/local/selectors/localSelectors.js#L93-L97): `LocalPlugin` needs to duplicate `hasNextSelector`, unchanged, just to use the modified `maxPageSelector`. There's actually less of that duplication in `LocalPlugin` than I expected, but I ask because I had run into a situation (pre-#718) where customizing [`sortedDataSelector`](https://github.com/GriddleGriddle/Griddle/blob/4bdc65e4c8b377751e9e2104ee4b53efa88e54f3/src/plugins/local/selectors/localSelectors.js#L106-L122) would have required basically cloning the entire plugin, due to cascading dependencies. Not sure if there's any way to defer composition of selectors?

joellanciaux:
Yeah, cascading dependencies are a bit of an issue at the moment. 

I believe @ryanlanciaux (who's currently in transit) can speak a little more about this, but I think he's been working on something that should handle some of the selector composition issues.

shorja:
@dahlbyk @joellanciaux @ryanlanciaux I'm also working on a solution to the problem of duplicated selectors in plugins, I'll post it in another PR since its not specifically related and might be a bit more of a dangerous change, should have that up sometime today.

src/components/TableBodyContainer.js
```diff
@@ -5,31 +5,33 @@ import compose from 'recompose/compose';
...
 )(({Row, visibleRowIds, style, className}) => (
   <OriginalComponent
-    rowIds={visibleRowIds}
-    Row={Row}
-    style={style}
-    className={className}
+  rowIds={visibleRowIds}
+  Row={Row}
+  style={style}
+  className={className}
```
dahlbyk: This component can pass-thru `props` directly, like the other components:

```javascript
)(props => <OriginalComponent {...props} />);
```

src/components/TableBodyContainer.js
```diff
@@ -5,31 +5,33 @@ import compose from 'recompose/compose';
...
+  connect(
+    (state, props) => ({
+      visibleRowIds: props.selectors.visibleRowIdsSelector(state),
+      className: props.selectors.classNamesForComponentSelector(state, 'TableBody'),
+      style: props.selectors.stylesForComponentSelector(state, 'TableBody'),
+    })
+  ),
   mapProps(props => {
-    const { components, ...otherProps } = props;
+    const { components, selectors, ...otherProps } = props;
```
dahlbyk: I noticed this is the only place you're pulling `selectors` out of `otherProps`. I think it's better this way, since `selectors` isn't meant to be exposed to the wrapped component - can you add repeat this change elsewhere?

src/components/TableBodyContainer.js
dahlbyk: Clarification: please rename `visibleRowIds` to `rowIds` so `props` can pass-thru directly.

Issue #733, opened 2017-09-01
[Discussion] Plugin Composition


dahlbyk:
## Griddle version

1.x and/or 2.x

## Expected Behavior

Plugins and components compose well.

## Actual Behavior

- Roles/relationships for `ContainerEnhancer`, `Container` and `Enhancer` are unclear
- Plugins can only replace individual components, so multiple plugins don't compose well
  - "One of the big issues I'm facing right now is composing a grid with multiple plugins." https://github.com/GriddleGriddle/Griddle/pull/731#issuecomment-326442141
- _What else?_

------------------------------------

Thoughts on how this situation can be improved?

dahlbyk:
For starters, we could accept plugins as functions (reducers, basically). The result of the function, called with the initial/current config, would be treated the same as a static plugin.

This may be sufficient to allow enhancer/container components to be composed instead of replaced.

dahlbyk:
@sel129 ran into this in https://github.com/GriddleGriddle/Griddle/issues/741:

> I have a situation where I have created 2 plugins. Both have a RowEnhancer. I expect that the RowEnhancer from both plugins will be applied to the Row Component.

It's reasonable to expect enhancers to build on each other, rather than be replaced, but changing that in 1.x would be a breaking change. Two ways occur to me to implement this without a breaking change:

1. Implement the plugin function idea above, which would allow a plugin to wrap the existing enhancer:

    ```javascript
    <Griddle
      plugins={[
        { components: { RowEnhancer: OC => props => <div>a <OC {...props} /></div> } },
        ctx => ({
          components: {
            RowEnhancer: OC => ctx.RowEnhancer(props => <div>b <OC {...props} /></div>)
          },
        }),
      ]} />
    ```

2. Update the enhancer composition logic to accept enhancers as an array of components, which would be composed with each other instead of replaced.

    ```javascript
    <Griddle
      plugins={[
        { components: { RowEnhancer: [OC => props => <div>a <OC {...props} /></div>] } },
        { components: { RowEnhancer: [OC => props => <div>b <OC {...props} /></div>] } },
      ]} />
    ```

Both of these would render `<div>a <div>b <Row {...props} /></div></div>`.

Between the two, I think I prefer the former, as it seems more deterministic.

dahlbyk:
FYI, I am hacking around this weekend on implementing support for plugins as functions.

shorja:
[...]

PR #743, opened 2017-09-26
Composable selectors

shorja:
## Griddle major version
1.9.0

This is a potential upgrade to the existing way selectors are created.

## Why these changes are made
Selectors are currently limited in how they can be overridden, this change would make it possible to override any number of selectors without having to change the dependency selectors. In combination with [this PR](https://github.com/GriddleGriddle/Griddle/pull/740) which changes all the built-in components to use the selectors on the context, this would allow for much more powerful overriding of selectors.

If this is a desirable feature, we need to have a discussion about how to make sure this isn't a breaking change, maybe this could be enabled by a flag on the Griddle component. Without a plugin using the new composableSelectors file, I believe this SHOULD be backwards compatible, would need to do some testing.

## Are there tests?
Not yet, needs comments first

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,7 +8,8 @@ import _ from 'lodash';
 import * as dataReducers from './reducers/dataReducer';
 import components from './components';
 import settingsComponentObjects from './settingsComponentObjects';
-import * as selectors from './selectors/dataSelectors';
+import * as baseSelectors from './selectors/dataSelectors';
+import * as composedSelectors from './selectors/composedSelectors';
 
 import { buildGriddleReducer, buildGriddleComponents } from './utils/compositionUtils';
 import { getColumnProperties } from './utils/columnUtils';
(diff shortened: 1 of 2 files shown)


shorja:
I will add some examples to storybook tomorrow to demonstrate how this works too

shorja:
selectors00

ryanlanciaux:
This is really great -- Going to spend some time going through this a bit but really promising and cool functionality! Thank you for adding this!

src/index.js
```diff
@@ -98,7 +99,206 @@ class Griddle extends Component {
 
     this.events = Object.assign({}, events, ...plugins.map(p => p.events));
 
-    this.selectors = plugins.reduce((combined, plugin) => ({ ...combined, ...plugin.selectors }), {...selectors});
```
dahlbyk: Before we dig into reviewing this code, can you pull your whole new block into `./utils/selectorUtils.js` to minimize the churn here? Something like:

```
this.selectors = flattenSelectors(selectors, plugins);
```

That should make it a lot easier to test your selector composition logic in isolation.

(Please squash this change, to avoid merge conflicts in this file.)

src/selectors/composedSelectors.js
```diff
@@ -0,0 +1,218 @@
+import Immutable from 'immutable';
+import { createSelector } from 'reselect';
+import _ from 'lodash';
+import MAX_SAFE_INTEGER from 'max-safe-integer'
+
+export const hasPreviousSelector = {
+  creator: ({currentPageSelector}) => {
```
dahlbyk: This may be too clever, but I wonder if this would be easier to write/read by attaching `dependencies` directly to the `creator` function:

```javascript
export const hasPreviousSelector = ({ currentPageSelector }) =>
  createSelector(
    currentPageSelector,
    currentPage => currentPage > 1);
hasPreviousSelector.dependencies = ["currentPageSelector"];
```

dahlbyk:
To clarify, the example story is a minimalist implementation of `LocalPlugin`?

Seeing your `griddleCreateSelector`, I have been toying with the idea of making it a drop-in replacement for `createSelector`. If it's given a selector as a dependency, it Just Works™️; if it's given a string, it resolves the dependency. Have a look at https://github.com/shorja/Griddle/compare/m-composable-selectors...dahlbyk:refactor-743; the first two commits are essentially a squashed reimplementation of this PR.

shorja:
Ok so I've been looking at your refactor, I really like the idea of a drop in replacement for createSelector that is able to adapt to whichever way it is called to replicate old behaviour. I have been working on a refactor to your refactor too. 

The way it was implemented with the global selectors, all of the import style references to selectors are recreated every time they are called which prevents the use of the auto memoization feature that reselect provides. I haven't quite finished my refactor but it should prevent this from being a problem and also the globalSelectors shouldn't be necessary.

However, I am concerned with the implications of changing the dataSelectors over to use the new dependency style. I'm worried that this could introduce some confusing behaviour for existing users especially since 'hard' imported selectors could now start being overridden without them knowing why. This was my goal with the original design of having a separate composedSelectors file, though this will still provide the overriding selector behaviour on the context which may also be a breaking change for existing users.

So, I'm going to propose making a change to the way the core components, selectors, reducers, etc... are used. I'm going to try moving this 'core' functionality over to a plugin. Griddle will always load one 'baseline' plugin, and by default it will always load the 'core' plugin. Then we can implement a new plugin, that has all of the same features of the 'core' plugin but will use all of the new composability features. Existing plugin users should still be able to leverage new composability features if they choose to use them but they won't be automagically opted in. This will also allow developers to totally override the 'baseline' with their own so that they don't have a cumbersome overriding process. This gets pretty close to Griddle being just a component building framework but is still backwards compatible.

I'll put up a PR as soon as I can get it done.

dahlbyk:
> I'm worried that this could introduce some confusing behaviour for existing users especially since 'hard' imported selectors could now start being overridden without them knowing why.

@ryanlanciaux and @joellanciaux, or other users, may disagree, but I'm not too concerned about people having taken hard dependencies on the selectors that have been exported thus far, because there hasn't been any way to inject dependencies until now. In other words, existing behavior can't change unless a new plugin/customization is added that uses this DI mechanism.

I do like the idea of thinking of the Griddle core behavior as a plugin; I am actually working on a branch that essentially takes that same view in refactoring how plugins are composed (#733). I will try to share that sooner rather than later, for your consideration.

shorja:
@dahlbyk My concern is specifically with changing the dataSelectors to using the quoted versions and their interaction with the core code.

For example if a user had defined a custom selector in their plugin that had the same name as a dataSelectors one, the current behaviour would be that it would be overridden in the context object but of course all of the hard references to it would stay the same whereas if the dataSelectors are created with the new logic using quoted dependencies, the 'hard' references will be overridden and all of the core code will now use that new selector. I'm more concerned about all of the code that dev's didn't write magically behaving differently. 

Though if we look at [PR 740](https://github.com/GriddleGriddle/Griddle/pull/740), adding the selectors to the context will also potentially cause unexpected behaviour for devs not familiar with any new composability features since that would ALSO make all the core code magically use any overridden selectors.

Sorry, I'm not trying to be contrarian, I'm just probably overly cautious about messing things up.

shorja:
@dahlbyk Oh and yeah I would love to see your approach to a plugin style refactor for the core code.

dahlbyk:
@shorja check out https://github.com/GriddleGriddle/Griddle/pull/757. The work is a few months old, so doesn't account for new functionality, but the key piece is that the initializer essentially accepts the default reducers, components, selectors, etc. in the shape of a plugin: https://github.com/dahlbyk/Griddle/blob/65ff192e3d8cef77a7975d372a3752fa3aa7b650/src/index.js#L69-L86.

mjosh954:
I too would like to see this functionality added!

PR #744, opened 2017-10-01
Render Loading component when data is null or undefined

dahlbyk:
## Griddle major version

1.x

1. Providing `data` of `null`/`undefined` no longer blows up with or without `LocalPlugin`.
2. When `data` is `null`/`undefined` we now render a `Loading` component instead of `NoResults`.

## Why these changes are made

User experience can be made less disruptive if the entire Griddle component doesn't need to be replaced by a loading animation (e.g. if a `Settings` component requires fetching new data).

griddle-loading

## Are there tests?

Stories!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,4 @@
-import { createStore, combineReducers, bindActionCreators, applyMiddleware } from 'redux';
+import { createStore, combineReducers, bindActionCreators, applyMiddleware, compose } from 'redux';
 import Immutable from 'immutable';
 import { createProvider } from 'react-redux';
 import React, { Component } from 'react';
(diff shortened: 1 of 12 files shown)


ryanlanciaux:
Awesome! this is super useful 👍

src/index.js
```diff
@@ -136,10 +136,13 @@ class Griddle extends Component {
       }
     );
 
+    const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose
```
andreme: @dahlbyk This breaks server side rendering because window is not available: ReferenceError: window is not defined

dahlbyk: Oh no! Should be easy to guard against with a `typeof window !== 'undefined'`, if you care to contribute?

andreme: https://github.com/GriddleGriddle/Griddle/pull/765

PR #745, opened 2017-10-03
[DOCS] Fix customization links in getting started docs

ibrahima:
The customization docs are not in the parent directory, they're in the same directory as this one. Currently the links to customization on [Getting Started](http://griddlegriddle.github.io/Griddle/docs/) are broken.

I suppose this could also be `Griddle/docs/customization` or just `customization`, not sure which is preferred.


dahlbyk:
Thanks! If you're feeling generous, please make the equivalent change in https://github.com/GriddleGriddle/griddle-docs; `gh-pages` here is generated from that repo.

ibrahima:
Huh, I was wondering if it was generated from something else, just didn't figure out where it was coming from.

I'm confused by this though; it looks like someone was already trying to fix it when they made it the broken thing: https://github.com/GriddleGriddle/griddle-docs/commit/1912cf49ec64758233691d2de6a74906797fcaf2 ...

PR #749, opened 2017-10-10
onSort with *correct* sortProperties

### LINKED ISSUE #716: onSort event not giving sortAscending ascending property

dahlbyk:
## Griddle major version

1.x

`onSort` is now fired just before the new `sortProperties` are dispatched.

This essentially renders `TableHeadingCell` obsolete, but deleting it is a potentially breaking change.

### Idea
What if events were fired from a Redux middleware instead of click handlers?

## Why these changes are made

Close #716; #717 was insufficient.

## Are there tests?

Existing story now demonstrates expected behavior.

It seems like a new React version changed behavior when `render()` returns a string, so the TS story has been adjusted.

### DIFF
--- a/src/components/TableHeadingCellEnhancer.js
+++ b/src/components/TableHeadingCellEnhancer.js
@@ -1,21 +1,4 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import compose from 'recompose/compose';
-import mapProps from 'recompose/mapProps';
-import getContext from 'recompose/getContext';
-import { combineHandlers } from '../utils/compositionUtils';
-
-const EnhancedHeadingCell = OriginalComponent => compose(
-  getContext({
-    events: PropTypes.object,
-  }),
-  mapProps(({ events: { onSort }, ...props }) => ({
-    ...props,
-    onClick: combineHandlers([
-      () => onSort && onSort(props.sortProperty || { id: props.columnId }),
-      props.onClick,
-    ]),
-  }))
-)(props => <OriginalComponent {...props} />);
+// Obsolete
+const EnhancedHeadingCell = OriginalComponent => OriginalComponent;
 
 export default EnhancedHeadingCell;
--- a/src/components/TableHeadingCellContainer.js
+++ b/src/components/TableHeadingCellContainer.js
@@ -4,9 +4,11 @@ import { connect } from '../utils/griddleConnect';
 import compose from 'recompose/compose';
 import mapProps from 'recompose/mapProps';
 import getContext from 'recompose/getContext';
-
+import withHandlers from 'recompose/withHandlers';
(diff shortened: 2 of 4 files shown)


src/components/TableHeadingCellEnhancer.js
```diff
@@ -1,21 +1,4 @@
...
-  mapProps(({ events: { onSort }, ...props }) => ({
-    ...props,
-    onClick: combineHandlers([
-      () => onSort && onSort(props.sortProperty || { id: props.columnId }),
-      props.onClick,
-    ]),
-  }))
-)(props => <OriginalComponent {...props} />);
+// Obsolete
+const EnhancedHeadingCell = OriginalComponent => OriginalComponent;
```
joellanciaux: Can this be removed?

dahlbyk: It can, but it's _technically_ a breaking change. Glad to remove if you're not concerned.

ryanlanciaux: Just to double confirm - we're leaving this here to avoid causing a potential breaking change right? I think we're maybe just outside the realm of where a 2.0 release makes sense. If I'm understanding correctly, we can leave this here for now to avoid breaking change and remove the obsolete field in a 2.0 version.

Thoughts?

dahlbyk: Correct. I left it as a no-op to be cleaned up for a 2.0 release.

The chances someone is actually depending on this seem quite low, but it's been exported as part of the public API.

dahlbyk:
On re-review I noticed a few gaps between core and local versions of `TableHeadingCellContainer`, so I've added a commit to close the gap (mostly from #699).

PR #751, opened 2017-10-10
Export connect

dahlbyk:
## Griddle major version

1.8+

1. Export `connect` directly, so it's more of a drop-in replacement for `react-redux`, i.e.

   ```diff
   -import { connect } from 'react-redux';
   +import { connect } from 'griddle-react';
   ```

2. Add static `storeKey` property, so default can be changed once app-wide

3. The `storeKey` story's `ConnectedComponent` was incorrectly using Griddle's `connect`; now it doesn't, and the demo works. (Bonus: demo's slightly prettier.)
   image

## Why these changes are made

I'm actually using this functionality now across an app, and it's a bit tedious.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -79,7 +79,7 @@ class Griddle extends Component {
       components:userComponents,
       renderProperties:userRenderProperties={},
       settingsComponentObjects:userSettingsComponentObjects,
-      storeKey = 'store',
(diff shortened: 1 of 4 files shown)


ryanlanciaux:
Thanks for adding this! Yea this is way nicer than needing to go several levels down to get griddle's connect 👍

PR #755, opened 2017-10-13
Component Re-render improvements

### LINKED ISSUE #750: Using customComponent: the component unmounts and re-mounts in every Griddle re-render
Hi, I have a table with two columns that are using custom components. One of them has a progress bar, whose intended purpose is to show the progress of an upload. When a file is being uploaded, Griddle re-renders accordingly every time there is a new progress and the progress bar updates. Now, I have a second column with a second custom component that is supposed to cancel the upload when an element inside it is clicked. However, this click is not being registered consistently because the component keeps mounting and unmounting at a rapid rate (because the upload progress updates quickly). I verified this by adding logs in the componentDidMount() method.
Is there a way to make Griddle not remount the cell or row components for every re-render?

joellanciaux:
## Griddle major version
v1

In an effort to resolve #750, I've made some updates to improve some of the diffing / `shouldComponentUpdate` logic in a handful of locations.

## Why these changes are made
- We didn't have a `shouldComponentUpdate` on the base Griddle, which lead to the root node being deleted on data updates when there might only be a small change.
- `GriddleConnect` was running into something similar where the custom store connected component is created in `render`, which leads to a new component when something changes.
- Reselect is awesome, but I think we made the assumption that the memoization took the properties into account, as we wanted to hang on to the previous `cellProperties` value for a particular column.
  - This isn't quite true, so we needed to make a selector factory so that the selector is scoped to _that_ column.
- ~Since we were setting custom cell value components to the `value` property, we were losing out on any potential render control within the custom component.  I moved this logic into `Cell` to give a little more control to the user.~
  - ~**This is a _slight_ API change** :warning:~

## Are there tests?
Covered by previous tests.

### DIFF
--- a/src/components/CellContainer.js
+++ b/src/components/CellContainer.js
@@ -10,7 +10,7 @@ import onlyUpdateForKeys from 'recompose/onlyUpdateForKeys';
 import {
   customComponentSelector,
   cellValueSelector,
-  cellPropertiesSelector,
+  cellPropertiesSelectorFactory,
   classNamesForComponentSelector,
   stylesForComponentSelector
 } from '../selectors/dataSelectors';
@@ -37,11 +37,9 @@ function getCellStyles(cellProperties, originalStyles) {
   return styles;
 }
 
-const ComposedCellContainer = OriginalComponent => compose(
-  getContext({
-    selectors: PropTypes.object,
-  }),
-  connect((state, props) => {
+const makeStateToProps = () => {
+  const cellPropertiesSelector = cellPropertiesSelectorFactory();
+  return (state, props) => {
     return {
       value: cellValueSelector(state, props),
       customComponent: customComponentSelector(state, props),
@@ -49,7 +47,11 @@ const ComposedCellContainer = OriginalComponent => compose(
       className: classNamesForComponentSelector(state, 'Cell'),
       style: stylesForComponentSelector(state, 'Cell'),
     };
-  }),
+  };
+}
+
+const ComposedCellContainer = OriginalComponent => compose(
+  connect(makeStateToProps),
   mapProps(props => {
     return ({
     ...props.cellProperties.extraData,
--- a/src/components/CellContainer.js
+++ b/src/components/CellContainer.js
@@ -4,8 +4,6 @@ import { connect } from '../utils/griddleConnect';
 import getContext from 'recompose/getContext';
 import mapProps from 'recompose/mapProps';
 import compose from 'recompose/compose';
-import shouldUpdate from 'recompose/shouldUpdate';
-import onlyUpdateForKeys from 'recompose/onlyUpdateForKeys';
 
 import {
   customComponentSelector,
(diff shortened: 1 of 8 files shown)


dahlbyk:
This is some great stuff!

> Updated GriddleConnect to not redefine connected component every render.

My first thought when I saw this `render()` method is that this could probably be a functional component using `withContext`. Am I missing something?

> Scoped selectors for cell properties.

I'm having a bit of trouble wrapping my head around this - can you talk me through it?

> This is a slight API change ⚠️

SemVer-wise, might consider reverting this change but including a story to demonstrate how to build a better Cell?

joellanciaux:
> My first thought when I saw this render() method is that this could probably be a functional component using withContext. Am I missing something?

I totally might be missing something, but the thought is to preserve the existing function signature of the `react-redux` `connect` function without overwriting the `store` context value for any custom components that might be using the vanilla `connect` to grab values from their application store.

If we're using `withContext` to change the `store` context value, we'll break regular `connect` for anything nested inside of Griddle.

> can you talk me through it?

This is explained pretty well in the [reselect documentation](https://github.com/reactjs/reselect#sharing-selectors-with-props-across-multiple-components). 

More or less, reselect is only memoizing the results of the `cellProperties` selector for a particular set of `state` and `props`.  Because we're using this selector for all of the `Cell` elements we're building, it's constantly destroying the memoized value as it's receiving new properties for each different `Cell`.

Scoping the selector's use keeps reselect's memoization meaningful.

> SemVer-wise, might consider reverting this change but including a story to demonstrate how to build a better Cell?

On further review, I think we can skip this change. Phew 😅

src/components/CellContainer.js
```diff
@@ -35,19 +37,21 @@ function getCellStyles(cellProperties, originalStyles) {
...
       cellProperties: cellPropertiesSelector(state, props),
       className: classNamesForComponentSelector(state, 'Cell'),
       style: stylesForComponentSelector(state, 'Cell'),
     };
-  }),
+  };
+}
+
+const ComposedCellContainer = OriginalComponent => compose(
+  connect(makeStateToProps),
```
dahlbyk: make? map?

src/components/CellContainer.js
```diff
@@ -4,11 +4,13 @@ import { connect } from '../utils/griddleConnect';
 import getContext from 'recompose/getContext';
 import mapProps from 'recompose/mapProps';
 import compose from 'recompose/compose';
+import shouldUpdate from 'recompose/shouldUpdate';
+import onlyUpdateForKeys from 'recompose/onlyUpdateForKeys';
```
dahlbyk: Both unused?

rdavidwu:
This is great stuff @joellanciaux 
@dahlbyk Do you know when this is expected to be released?

ryanlanciaux:
This one is in npm right now but not as latest. I'm going to give it another run through and push as latest sometime this weekend :+1:

PR #756, opened 2017-10-13
Preserve Sort Properties

### LINKED ISSUE #697: sortProperties override users sorting on data change
## Griddle version
1.6.0

## Expected Behavior
When changing the data prop on the data, the users sorting selection should remain

## Actual Behavior
From the docs: "The sortProperties prop defines the default sort options for Griddle. " implying that the sortProperties should not get reapplied when changing other props.

## Steps to reproduce
* Specify the following on a table sortProperties:[ { id:"name" } ]}
* Sort the table on another column
* Change the value of any other prop on the Griddle element
* Observe that the default sort gets reapplied.

The same also happens with pageProperties

joellanciaux:
## Griddle major version
v1

Updates `componentShouldUpdate` to only pass properties that have changed to the `updateState` action.

Previously, if an initial sort value was passed in while data changed, the user-selected sorting option would be reset to the initial value.

## Why these changes are made
To resolve #697

## Are there tests?
No 😞

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -156,9 +156,17 @@ class Griddle extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const { data, pageProperties, sortProperties } = nextProps;
+    const newState = Object.keys(nextProps).reduce((updatedState, key) => {
+      return this.props[key] !== nextProps[key] ? {
+      ...updatedState,
+      [key]: nextProps[key],
+     } : updatedState;
+    }, {});
 
-    this.store.dispatch(actions.updateState({ data, pageProperties, sortProperties }));
+    // Only update the state if something has changed.
+    if (Object.keys(newState).length > 0) {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -156,9 +156,17 @@ class Griddle extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const { data, pageProperties, sortProperties } = nextProps;
-
-    this.store.dispatch(actions.updateState({ data, pageProperties, sortProperties }));
+    const newState = Object.keys(nextProps).reduce((updatedState, key) => {
+      return this.props[key] !== nextProps[key] ? {
+      ...updatedState,
+      [key]: nextProps[key],
```
dahlbyk: Mutate accumulator here for performance?

joellanciaux: Good call! Updated to `_.pickBy`

PR #757, opened 2017-10-16
Extract Initializer

dahlbyk:
## Griddle major version

1.x

Pulls the default+plugin+props composition out of the Griddle constructor into a testable initializer.

This is really just prep to work on supporting function plugins for #733.

## Why these changes are made

Tests are good, especially for this sort of composition (which seems likely to get more complex).

## Are there tests?

Lots!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,16 +10,8 @@ import settingsComponentObjects from './settingsComponentObjects';
 import * as selectors from './selectors/dataSelectors';
 
 import init from './utils/initializer';
-import { setSortProperties } from './utils/sortUtils';
 import { StoreListener } from './utils/listenerUtils';
 import * as actions from './actions';
-
-const defaultEvents = {
-  ...actions,
-  onFilter: actions.setFilter,
-  setSortProperties
-};
-
 
 const defaultStyleConfig = {
   icons: {
--- a/src/index.js
+++ b/src/index.js
@@ -4,45 +4,10 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
 
-import * as dataReducers from './reducers/dataReducer';
-import components from './components';
-import settingsComponentObjects from './settingsComponentObjects';
-import * as selectors from './selectors/dataSelectors';
-
+import corePlugin from './core';
 import init from './utils/initializer';
 import { StoreListener } from './utils/listenerUtils';
 import * as actions from './actions';
-
-const defaultStyleConfig = {
-  icons: {
-    TableHeadingCell: {
-      sortDescendingIcon: '▼',
-      sortAscendingIcon: '▲'
-    },
-  },
-  classNames: {
-    Cell: 'griddle-cell',
-    Filter: 'griddle-filter',
-    Loading: 'griddle-loadingResults',
-    NextButton: 'griddle-next-button',
-    NoResults: 'griddle-noResults',
-    PageDropdown: 'griddle-page-select',
-    Pagination: 'griddle-pagination',
-    PreviousButton: 'griddle-previous-button',
-    Row: 'griddle-row',
-    RowDefinition: 'griddle-row-definition',
-    Settings: 'griddle-settings',
-    SettingsToggle: 'griddle-settings-toggle',
-    Table: 'griddle-table',
-    TableBody: 'griddle-table-body',
-    TableHeading: 'griddle-table-heading',
-    TableHeadingCell: 'griddle-table-heading-cell',
-    TableHeadingCellAscending: 'griddle-heading-ascending',
-    TableHeadingCellDescending: 'griddle-heading-descending',
-  },
-  styles: {
-  }
-};
 
 class Griddle extends Component {
   static childContextTypes = {
@@ -58,27 +23,11 @@ class Griddle extends Component {
     super(props);
 
     const {
+      core = corePlugin,
       storeKey = Griddle.storeKey || 'store',
     } = props;
 
-    const { initialState, reducers, reduxMiddleware } = init.call(this, {
-      reducers: dataReducers,
-      components,
-      settingsComponentObjects,
-      selectors,
-      styleConfig: defaultStyleConfig,
-
-      pageProperties: {
-        currentPage: 1,
-        pageSize: 10
-      },
-      enableSettings: true,
-      textProperties: {
-        next: 'Next',
-        previous: 'Previous',
-        settingsToggle: 'Settings'
-      },
-    });
+    const { initialState, reducers, reduxMiddleware } = init.call(this, core);
 
     const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose
     this.store = createStore(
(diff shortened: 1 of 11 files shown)


src/utils/__tests__/initilizerTests.js
```diff
@@ -0,0 +1,362 @@
...
+    props: {
+      plugins: [
+        { events: { Plugin: 0, User: false } },
+        { events: { Plugin: 1 } },
+      ],
+      events: { User: true, User2: true },
+    },
+  };
+  const defaults = {
+    // TODO: bug that defaultEvents is not used?
```
dahlbyk: @joellanciaux unclear from history, but I'm pretty sure you created `defaultEvents` but it was never used. Thoughts?

joellanciaux: Yikes.  I'm having a hard time following what was happening going off the `blame`. 😅 

I _think_ it was related to some composition updates I was wanting at the time, but I'm positive it's a perfect candidate for removal.

src/index.js
```diff
@@ -67,76 +63,32 @@ class Griddle extends Component {
...
-    this.settingsComponentObjects = Object.assign({}, settingsComponentObjects, ...plugins.map(p => p.settingsComponentObjects), userSettingsComponentObjects);
-
-    this.events = Object.assign({}, events, ...plugins.map(p => p.events));
-
-    this.selectors = plugins.reduce((combined, plugin) => ({ ...combined, ...plugin.selectors }), {...selectors});
-
-    const mergedStyleConfig = _.merge({}, defaultStyleConfig, ...plugins.map(p => p.styleConfig), styleConfig);
-
-    const pageProperties = Object.assign({}, {
+      pageProperties: {
```
joellanciaux: It looks like the `pageProperties` and `sortProperties` properties were removed. Can this be re-added?

dahlbyk: As far as I could/can tell, there's actually nothing special about `pageProperties` and `sortProperties` relative to any other "extra" properties that are treated as `initialState`: note tests don't change in https://github.com/GriddleGriddle/Griddle/pull/757/commits/d69275ab85df6a6c60d17ae1818e5e9a45d4612a and https://github.com/GriddleGriddle/Griddle/pull/757/commits/0cbc7d8e14ff64d84f160734885824d4f0c69bbf.

dahlbyk:
> I _am_ seeing some issues on the TS front in storybook that I'm having a hard time tracking down.  I might need to lean on @ryanlanciaux to investigate that bit. Are you seeing anything?

I don't recall any issues from a month ago when I was working on this, but I can take another look.

I take it you approve of this direction? If so, I'll rebase on `master` to account for evolution since then.

src/utils/__tests__/initilizerTests.js
```diff
@@ -0,0 +1,388 @@
...
+    },
+  };
+  const defaults = {};
+
+  const res = init.call(ctx, defaults);
+  assert.truthy(res);
+
+  assert.deepEqual(ctx.events, {
+    Plugin: 1,
+    User: false, // TODO: bug that plugins overwrite user events?
```
dahlbyk: @joellanciaux @ryanlanciaux is it intentional that `events` provided in `<Griddle events={...}>` are overwritten by `events` from plugins?

src/utils/__tests__/initilizerTests.js
```diff
@@ -0,0 +1,388 @@
...
+  };
+  const defaults = {};
+
+  const res = init.call(ctx, defaults);
+  assert.truthy(res);
+  assert.truthy(res);
+
+  assert.false('defaults' in ctx.listeners);
+  assert.deepEqual(ctx.listeners.plugin(), 1);
+  assert.deepEqual(ctx.listeners.user(), false); // TODO: bug that plugins overwrite user listeners?
```
dahlbyk: @joellanciaux @ryanlanciaux again: should `<Griddle listeners={...}>` be overwritten by plugins?

cc @shorja

src/utils/__tests__/initilizerTests.js
dahlbyk: ✅ https://github.com/GriddleGriddle/Griddle/pull/757/commits/764a47edddd302a6f66840f7dd5f1190f5493c45

dahlbyk:
Inspired by #760, I've added commits to pull the current `defaults` into a `CorePlugin` of sorts that can be replaced (or omitted) via `<Griddle core={...} />`. Not sure if `core` is the right name for it, but it does seem a useful extra seem to adjust Griddle default behavior.

I did not bring in @shorja's `core` support for accepting a function from `props`, so parsing `renderProperties` from `children` (`RowDefinition`/`ColumnDefinition`) is the only behavior that's really not extensible. I'm not thrilled with `core` as function-of-`props` because for #733 I have been thinking of plugins as more of a reducer. But I'm not sure what a better pattern for handling `renderProperties` would be.

shorja:
@dahlbyk Yeah now that I look at the core plugin taking in props, that's probably not a great way to handle stuff, especially since the pattern seems to be passing in init config to a plugin is done outside griddle -> <Griddle plugins={[SomePlugin({...someConfig}] />. So really anything that should go into core really should be passed in like that, props should have nothing to do with it.

Is the idea of plugins as reducers to basically chain the output of built plugins into the next plugin to build and then each one can do what it wants to wrap components? I've also been thinking about refactoring the component composition logic, included in my mega PR was a change to allow enhancers to 'stack' but I was planning on revisiting it. I drew up a diagram that explains the behaviour I was thinking about.

componentcomposition00

dahlbyk:
> Is the idea of plugins as reducers to basically chain the output of built plugins into the next plugin to build and then each one can do what it wants to wrap components? I've also been thinking about refactoring the component composition logic, included in my mega PR was a change to allow enhancers to 'stack' but I was planning on revisiting it.

Yeah, given a series of function plugins, you can think of `initializer` as essentially `plugins.reduce((config, plugin) => plugin(config), core)`. My first commit in this direction was https://github.com/GriddleGriddle/Griddle/commit/e3f9481af3693160e745ff0f312f998722cbca34, but I wanted to land this initial refactoring first.

I like the idea of being able to apply multiple enhancers, as your diagram explains, but explaining "enhancers cleared by a new component" rules might be less intuitive than making (function) plugins decide how old/new enhancers should compose (c.f. https://github.com/GriddleGriddle/Griddle/issues/733#issuecomment-330975239). I'd suggest the plugin composition conversation continue over there.

ryanlanciaux:
I like this direction and definitely wonder if pushing the enhancer registration logic to the plugins themselves would make the logic easier to understand for anyone using the library.

dahlbyk:
> > I am seeing some issues on the TS front in storybook that I'm having a hard time tracking down. I might need to lean on @ryanlanciaux to investigate that bit. Are you seeing anything?
>
> I don't recall any issues from a month ago when I was working on this, but I can take another look.

@joellanciaux are you still seeing this? Thoughts on the `pageProperties`/`sortProperties` change?

ryanlanciaux:
I'm in favor of moving this one forward. If there are no storybook + typescript issues (I'm not seeing any atm) I will merge this. Sound good? :D

PR #759, opened 2017-10-18
Updated sort to allow sorting on nested columns

joellanciaux:
## Griddle major version
v1

Fixes a bug that prevents sorting on nested columns (ex: `location.state`)

## Why these changes are made
To fix the bug!

## Are there tests?
Yep!


ryanlanciaux:
👍

PR #760, opened 2017-10-19
M major refactor

shorja:
## Griddle major version
1.10.0

A lot, this is sort of an 'omnibus' PR that includes all of the Griddle refactor stuff I've been working on. Unfortunately its gotten to the point where this stuff makes most sense all being together. This is not meant to be merged, but to be a place to centralize the conversation about the changes I'm proposing and also makes it easier to run them all at once. This branch is currently very active so expect frequent updates. Also, if this is not a good way of approaching these changes please let me know.

- Griddle now only uses plugins to build the table. I have moved all of the table code in src/ into src/plugins/core, I have been referring to this stuff as the 'core' plugin. By default this 'core' plugin is used as the baseline to start building Griddle but now you can pass in a prop the is a different baseline plugin. This is a big change to the Griddle component itself, the basic idea though is that there shouldn't be any dependencies on any actual table code when building the plugins, we should just get components, reducers, selectors, actions, and store state from each plugin and layer those on top of each other. My intention is to make Griddle at its core no longer a table library, but a component building framework.

- I'm trying to move everything any individual component might depend on to the context. This improves the already great behaviour overriding features Griddle already has since it creates a single point of truth for the props that are mapped into the components.

- Selectors are probably the single biggest change I've been working on, and I still have more improvements I want to make here. As has been discussed in [this PR](https://github.com/GriddleGriddle/Griddle/pull/743) selectors are now individually overridable via the use of a special griddleCreateSelector function and composeSelectors utility function. I have added some improvements on that pull request here: griddleCreateSelector now supports 'mixed mode' selector creation where you can use a combination of literal selector function and string selector dependencies. It now also will override every single instance of a selector being defined so imported selectors now automagically get the proper overridden behaviour. Selectors now also have a factory prop on them so you can get individual instances of each selector. The factory also accepts an object that allows you to redefine the dependency selectors for every run of the factory. I have improvements coming soon to allow very easy factory-ified creating of selectors for components that are not rendered just once, rows and cells being the most common use cases.

- Actions have also been moved to the context. This object is currently very simple, just all of the actions defined in the plugins. I would like to take a closer look at this and see how overriding should be handled.

- Multiple Enhancers can now be applied to a given component.
## Why these changes are made
Generally to support improving the plugin and overriding behaviour of Griddle overall. The table we are building has a ton of features which can be mixed and matched and so making the infrastructure really really solid is our primary motivation. 

## Are there tests?
Testing is ongoing, I have been adding tests for the selector composition utility, and I have updated existing tests to function properly with this refactor.


dahlbyk:
Related: https://github.com/GriddleGriddle/Griddle/pull/757#issuecomment-338301964

PR #765, opened 2017-11-06
Check for existence of window before using it

andreme:
## Griddle major version
1.x

## Why these changes are made
Fix server side rendering where window is not available

## Are there tests?
Either there are no test that cover this area or the test runner creates a window object


dahlbyk:
Thanks!

Issue #767, opened 2017-11-09
inputbox in customFilterComponent lose focus


radziksh:
in fact this is a repetition of another issue, which was closed without a solution:
https://github.com/GriddleGriddle/Griddle/issues/475

## Griddle version
1.9.0
## Expected Behavior
input should not lose focus.

## Actual Behavior
input loses focus

## Steps to reproduce

```typescript 

export class SearchStringFilter extends React.Component<TProps, {}> {

  render() {

    return (
      <input
        value={this.props.searchString}
        onChange={this.props.onChange} //dispatch("SOME_ACTION_FOR_CHANGE_VALUE") 
      /> 
    );
  }
}

<Griddle
    components={{
        Filter: SomeCustomFilter
    }}
```
I use my own redux store and every time I send a new searchString to onChange event. But unfortunately the focus of the filter is lost. What could be the problem? maybe there is some kind of workaround?

note:  I've read that re-initializing HOC inside a render function often leads to this problem, but that's just my guess. 

   > Don’t Use HOCs Inside the render Method
React’s diffing algorithm (called reconciliation) uses component identity to determine whether it should update the existing subtree or throw it away and mount a new one. If the component returned from render is identical (===) to the component from the previous render, React recursively updates the subtree by diffing it with the new one. If they’re not equal, the previous subtree is unmounted completely.
https://reactjs.org/docs/higher-order-components.html

I tried to find a problem with HOC in griddle but did not find it. At first sight everything is all right.

dahlbyk:
As you note, losing focus usually means the `input` re-rendered. Your `SearchStringFilter` seems fine, but would depend on `searchString` and `onChange`. You mention your own Redux store - can you provide a more complete repro that includes that interaction? Our Storybook provides a great environment for repros, and includes an [example of connecting to another store](https://github.com/GriddleGriddle/Griddle/blob/3466e7cc19f0f5fba591dcf487ec7d5abec39e86/stories/index.tsx#L844-L893).

radziksh:
@dahlbyk 
```typescript
import { getContext, compose, mapProps } from "recompose";
const CustomLayoutContainer = OriginalComponent => compose(
    getContext({
        components: React.PropTypes.object,
    }),
    griddleUtils.connect(
        (state, props: any) => {

            return ({
                gridState: props.components.gridState,
                onSearchChange: props.components.onSearchChange,
                className: selectors.classNamesForComponentSelector(state, "Layout"),
[...]

PR #768, opened 2017-11-14
Add external filter (via Redux) example

dahlbyk:
## Griddle major version

1.x

Adds story to troubleshoot #767

@radziksh is this a reasonably faithful reproduction of your situation? I improvised `SomePageConnected`, but it seems to be approximately what you're trying to achieve. As implemented here, I'm able to filter data from either `input` without losing focus. Thoughts?

### DIFF
--- a/stories/index.tsx
+++ b/stories/index.tsx
@@ -916,10 +916,15 @@ storiesOf('Griddle main', module)
       })
     )(CustomFilterComponent);
 
+    const plugins = [
+      LocalPlugin,
+      {
+        components: { Filter: CustomFilterConnectedComponent },
+      },
+    ];
     const SomePage = props => (
       <div>
-        <Griddle data={props.data} plugins={[LocalPlugin]} storeKey="griddleStore"
-          components={{ Filter: CustomFilterConnectedComponent }} />
+        <Griddle data={props.data} plugins={plugins} storeKey="griddleStore" />
 
         Component outside of Griddle that's sharing state
         <CustomFilterConnectedComponent />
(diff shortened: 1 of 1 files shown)


radziksh:
@dahlbyk I'm not sure (so happens in my particular case), but I think you should change:
```
const SomePage = props => (
      <div>
        <Griddle data={props.data} plugins={[LocalPlugin]} storeKey="griddleStore"
          components={{ Filter: CustomFilterConnectedComponent }} />

        Component outside of Griddle that's sharing state
        <CustomFilterConnectedComponent />
     </div>
    );
```
to:
``` 

const  components={ Filter: CustomFilterConnectedComponent };

const SomePage = props => (
      <div>
        <Griddle data={props.data} plugins={[LocalPlugin]} storeKey="griddleStore"
          components={components} />

        Component outside of Griddle that's sharing state
        <CustomFilterConnectedComponent />
     </div>
    );
```
since previous code will result in resetting the state of the grid if some action is not directly related to the grid. Since the grid will think that each time it receives a new state.

dahlbyk:
@radziksh thanks for the feedback. As you note, it shouldn't impact Griddle 1.10 or newer, but there might be issues in earlier versions.

radziksh:
@dahlbyk in fact, this behavior remains on version 1.10.1.
you can reproduce it very simply:
1) create a test action that will be dispatched 1 time per second.
2) transfer the components through an anonymous object:
 ```components = {{Filter: CustomFilterConnectedComponent}}```
3) see that griddle will reset state. ```(Object.keys(newState).length``` in ```componentWillReceiveProps``` will return ```1``` all the time.

Issue #774, opened 2017-11-24
Filter placeholder and numeric sorting


ronwillemse:
## Griddle version
1.9.0

Hi, is there an option to change the filter placeholdertext to something like search without writing a custom filter component? I think it is much overdone to add a custom component only to change the placeholder text.

Second thing is the way of sorting. For example if i have numbers it sorts like:

1
10
100
1000
2
3

instead of

1
2
3
10
...

How can I solve this?

dahlbyk:
> is there an option to change the filter placeholdertext to something like search without writing a custom filter component? I think it is much overdone to add a custom component only to change the placeholder text.

Not currently, but this should be a relatively easy patch if you want to submit a PR:

1. Add `filterPlaceholder` to [`initialState.textProperties`](https://github.com/GriddleGriddle/Griddle/blob/1e6346ab37df96adcd69a22ef2959f7be2f5db00/src/core/initialState.js#L40-L44)
2. Use `textSelector` in `FilterContainer` to add that value to `props` (e.g. [`PreviousButtonContainer`](https://github.com/GriddleGriddle/Griddle/blob/1e6346ab37df96adcd69a22ef2959f7be2f5db00/src/components/PreviousButtonContainer.js#L6)
3. Use the new `props` value in `Filter`

Once connected to `textProperties`, you can override the text [from a plugin](https://github.com/GriddleGriddle/Griddle/blob/1e6346ab37df96adcd69a22ef2959f7be2f5db00/stories/index.tsx#L1679-L1684) or [directly](https://github.com/GriddleGriddle/Griddle/blob/1e6346ab37df96adcd69a22ef2959f7be2f5db00/stories/index.tsx#L1535).

> Second thing is the way of sorting. For example if i have numbers it sorts like...

You can provide a custom `sortMethod` either for the entire grid or specific columns. Check out how [this custom sort function](https://github.com/GriddleGriddle/Griddle/blob/1e6346ab37df96adcd69a22ef2959f7be2f5db00/stories/index.tsx#L27-L42) is used in the Storybook.

ronwillemse:
If it is possible to change the textproperties for next and previous than it should also be possible to change other values like the placeholder. Let me try to find a workaround with te use of textSelector

dahlbyk:
[...]

PR #780, opened 2017-12-21
made it possible to change filter placeholder text

miguelsaldivar:
## Griddle major version
1.11.0

Fixed one of the issues regarding #774, which is an easy change to filter place holder text

## Why these changes are made
To be able to change the place holder text without creating a new component. 

## Are there tests?
Ran `npm run test` and it passed all tests.


stories/index.tsx
```diff
@@ -767,6 +767,14 @@ storiesOf('Griddle main', module)
       </Griddle>
     )
   })
+  .add('with Filter place-holder', () => {
```
dahlbyk: Can you move this down into a new Filter section of stories, similar to [the story for `settingsToggle`](https://github.com/GriddleGriddle/Griddle/blob/a3ace3027472caff209428cb30fb792bce28163f/stories/index.tsx#L1538-L1544)?

src/core/initialState.js
```diff
@@ -40,6 +40,7 @@ export default {
   textProperties: {
     next: 'Next',
     previous: 'Previous',
-    settingsToggle: 'Settings'
+    settingsToggle: 'Settings',
+    filterPlaceHolder: 'Filter'
```
dahlbyk: 1. Can we rename this as `filterPlaceholder` since the attribute is `placeholder` not `placeHolder`?
2. Maybe keep the list alphabetized?

src/components/Filter.js
```diff
@@ -5,7 +5,8 @@ class Filter extends Component {
   static propTypes = {
     setFilter: PropTypes.func,
     style: PropTypes.object,
-    className: PropTypes.string
+    className: PropTypes.string,
+    text: PropTypes.string
```
dahlbyk: Would it make sense to rename this prop as `placeholder`, to align with the `textProperties` key and to make explicit how the value will be used?

Bonus points if you add a trailing comma.

dahlbyk:
@ryanlanciaux also, this includes a `package-lock.json` file. Have you specifically been avoiding including one (in which case we should add to `.gitignore`?

stories/index.tsx
miguelsaldivar: Should I also move all the filter examples from plugins, to this section?

dahlbyk:
Changes look good to me. I'm ambivalent on relocating the other Filter story, as it really belongs in a Redux section.

@ryanlanciaux?

miguelsaldivar:
@dahlbyk @ryanlanciaux I don't think that's a bad idea to put it into a redux section, along with the custom column chooser within the settings story.

miguelsaldivar:
Were these changes good, or should I fix them? :smile_cat:

dahlbyk:
Sorry for the delay, @miguelsaldivar. All the maintainers are stretched pretty thin, but we do appreciate the contribution!

miguelsaldivar:
No no, it's fine don't even worry :smile:, I'll look for some other bugs I can fix to help you guys out!

PR #800, opened 2018-03-03
Stop filter from matching invisible properties

### LINKED ISSUE #725: how to stop filtering on not visible field?
## Expected Behavior
"filterable" is working as expected on visible field.

## Actual Behavior
But In actually filtering is also filtering data from hidden field as well. so if we something type in search filter and it will return result as well still it is not matching in visible filterable columns.

## Steps to reproduce
1.You can see this in storybook example page given in griddle
It is under **with local and filterable set** example.

2. Just type "unit" in searchbox above grid.

3. It will return you three result.
screen shot 2017-08-22 at 2 54 53 pm
You can see in above image that none of record and none of there field contents **unit** but actually **unit** is exits in **country** field which is not visible.

As i remember in griddle version 0.x has same behaviour. how i can stop filtering on hidden fields in grid(version 1.x).

mbland:
## Griddle major version
1

Closes #725. This fixes the problem of invisible fields getting matched by the filter by updating the filteredDataSelector() condition to exclude griddleKey and any values not covered by a ColumnDefinition if any ColumnDefinitions are present.

Bear in mind that columns corresponding to other React components may still match the filter based on generated HTML properties rendered as part of the value.toString().toLowerCase() expression. At this point the best course of action may be to recommend that such ColumnDefinitions are tagged with `filterable={false}` if it's a significant problem for the user.

## Why these changes are made

The previous `if (filterable === false)` condition meant that `filterable={false}` must be explicitly set on a ColumnDefinition in order for any property returned by row.keySeq() to be ignored. This resulted in the filter matching invisible row properties such as griddleKey and any properties in the data not covered by a ColumnDefinition.

I confirmed this by locally changing the expression at the end of this filter from:

```javascript
  return value &&
    value.toString().toLowerCase().indexOf(filterToLower) > -1;
```

to:

```javascript
  if (value && value.toString().toLowerCase().indexOf(filterToLower) > -1) {
    console.log('KEY:', key, 'VALUE:', value.toString().toLowerCase());
    return true;
  }
  return false;
```

and observed the unexpected filter matches on invisible row properties in the console.

## Are there tests?

I've added updated existing tests, added new tests, and validated the filter behavior in the Storybook.

### DIFF
--- a/src/plugins/local/selectors/__tests__/localSelectorsTest.js
+++ b/src/plugins/local/selectors/__tests__/localSelectorsTest.js
@@ -249,18 +249,52 @@ test('filteredDataSelector filters data respecting filterable', test => {
         name: {
           filterable: false,
         },
+        weapon: {
+          filterable: true,
+        }
       }
     },
     filter: 'H',
     data: [
       { id: '1', name: 'luke skywalker', weapon: 'light saber' },
-      { id: '2', name: 'han solo', weapon: 'blaster' }
+      { id: '2', name: 'han solo', weapon: 'blaster' },
     ]
   });
 
   test.deepEqual(selectors.filteredDataSelector(state).toJSON(), [
     { id: '1', name: 'luke skywalker', weapon: 'light saber' }
   ]);
+});
+
+test('filteredDataSelector matches ColumnDefinition fields only', test => {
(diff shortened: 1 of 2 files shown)


dahlbyk:
💯

PR #808, opened 2018-04-16
Release 1.13.0

dahlbyk:
Trying to ship 1.13.0 with latest changes, but working through an npm issue.

Per recent Gitter conversation, adding a `prepare` step to build when installed from GitHub. @chrisortman, can you verify this works as expected if you install `GriddleGriddle/Griddle#v1.13.0`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "griddle-react",
-  "version": "1.12.0",
+  "version": "1.13.0",
   "description": "A fast and flexible grid component for React",
   "keywords": [
     "react-component",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,10 @@
     "build-ts": "cp src/module.d.ts dist/module/",
     "build-umd": "webpack --config webpack.config.js",
     "build-modules": "cross-env BABEL_ENV=build babel src --out-dir dist/module ",
-    "ship-it": "npm run build && npm publish --tag next"
+    "postpublish": "git push --tags",
+    "prepare": "npm run build",
+    "preversion": "npm test",
+    "ship-it": "npm publish --tag next"
   },
   "peerDependencies": {
     "react": ">=15"


chrisortman:
With yarn 1.5.1 it does not, however I think there may be a bug out there related to this behavior being broken. I am going to test with yarn 1.6.0 and NPM and report back.

chrisortman:
Using 'GriddleReact/Griddle#1.13.0' as the package identifier:

- yarn 1.5.1 : No
- yarn 1.6.0 : No
- npm 5.6.0 : Yes

I think this is the relevant yarn issue https://github.com/yarnpkg/yarn/issues/5235

That yarn issue says that the broken behavior is only triggered with certain formats of package name. I have tested

- https://github.com/GriddleGriddle/Griddle.git#v1.13.0 : No
- git@github.com:GriddleGriddle/Griddle.git#v1.13.0 : No

I then started thinking that maybe yarn would run these scripts when it puts something into its cache, not takes it out, so I did a `yarn cache clean` and tried installing

After cache clean

- GriddleGriddle/Griddle : No
- https://github.com/GriddleGriddle/Griddle.git : New error

```
An unexpected error occurred: "https://registry.yarnpkg.com/core-js/-/core-js-1.2.7.tgz: EEXIST: file already exists, mkdir '/Users/me/Library/Caches/Yarn/v1/npm-core-js-1.2.7-652294c14651db28fa93bd2d5ff2983a4f08c636/library/fn'".
yarn install  32.91s user 20.18s system 128% cpu 41.457 total
```

Seems surprising to me the yarn bug has existed for as long as it has

dahlbyk:
@chrisortman thanks!

@joellanciaux @ryanlanciaux `@1.3.0` has been published to npm and tagged as `next`. Not sure how you evaluate fitness to bump `latest`, so I'll leave that to you.

dahlbyk:
Well I merged this so #813 could take a dependency on the `prepare` script, but CodeSandbox doesn't seem to support that yet. Sigh.

Still open to feedback on the script changes.

PR #753, opened 2017-10-11
v0: fix React.createClass and PropTypes for react 16

vakopian:
## Griddle major version
v0 (0.8.1)

## Changes proposed in this pull request
use `createReactClass` from `create-react-class` package and `PropTypes` from `prop-types` package.

## Why these changes are made
Make it compatible with React 16

PR #809, opened 2018-04-16
Release 0.8.2

dahlbyk:
This PR aligns the build process with v1 as of #808.

I have published 0.8.2, to release https://github.com/GriddleGriddle/Griddle/pull/753, but the build is failing due to `react-addons-test-utils` dependency hell.

@vakopian do you have any ideas how to resolve the [failure](https://travis-ci.org/GriddleGriddle/Griddle/builds/362748586) due to `react-addons-test-utils` shift from `react` to `react-dom` peer dependency?

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,9 +49,13 @@
     "webpack-dev-server": "1.12.1"
   },
   "scripts": {
+    "grunt": "grunt",
     "test": "karma start --browsers PhantomJS",
     "test-single": "karma start --singleRun true --browsers PhantomJS",
-    "grunt": "grunt"
+    "postpublish": "git push --tags",
+    "prepare": "grunt build",
+    "preversion": "npm test",
+    "ship-it": "npm publish --tag version0"
   },
   "repository": {
     "type": "git",


thathenderson:
The path to import/require React Test Utils is now `react-dom/test-utils` instead of `react-dom/lib/ReactTestUtils`. The tests would have to update all those references.

dahlbyk:
@thathenderson a PR into this branch is probably easiest, if you don't mind.

I have never dealt with targeting multiple React versions, so my biggest concern is making sure stuff keeps working with older version (especially for v0.x). Or am I being overly paranoid about `devDependencies`?

thathenderson:
@dahlbyk You know, I was wondering about that too. As of React 15.5, Test Utils and testutils.createRenderer were [moved inside react-dom](https://reactjs.org/blog/2017/04/07/react-v15.5.0.html#react-test-utils), which means the apis the tests use need to change. 15.5 would have to be the lowest supported React/ReactDom versions.

thathenderson:
Does v0 still need to offer support for React 15.4 and below (including 0.14.x)?

dahlbyk:
> Does v0 still need to offer support for React 15.4 and below (including 0.14.x)?

I need @joellanciaux and/or @ryanlanciaux to weigh in on this. I'm mostly here for 1.0+.

> I guess we could have different react/react-dom versions in devDependencies than in peerDependencies? Is that bad practice?

A bit unconventional, perhaps, but I can't see much way around it if one intends wide support. It does lower one's confidence in tests; realistically tests should be run for React 14 and 15 and 16, but I have no idea how to pull that off.

thathenderson:
> It does lower one's confidence in tests; realistically tests should be run for React 14 and 15 and 16, but I have no idea how to pull that off.

Agreed. I'm not sure there's a way to continue supporting an api that's multiple deprecations behind.

PR #811, opened 2018-04-17
Release 0.8.2 -- Fixed Tests

thathenderson:
Note: went ahead and made the react/react-dom 0.14.x and 15.4.x devDependencies changes discussed briefly in [#809](https://github.com/GriddleGriddle/Griddle/pull/809). I can easily revert that if anyone is uncomfortable with the idea, but then a discussion needs to be had around how to fix `react-addons-test-utils` issues.

## Griddle major version
0.8.2 (fixes targeted for this version release)

* Deprecating React 0.14.x and 15.4.x support in devDependencies
* Fixing React Test Utils package dependency, along with related dependencies (ShallowRenderer)
* Fixing es2015 Map constructor errors with a babel polyfill

## Why these changes are made
To get passing tests!

## Are there tests?
No new tests; all existing tests pass (as of this branch)

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -5,6 +5,7 @@ module.exports = function(config) {
     basePath: '',
     frameworks: ['jasmine'],
     files: [
+      './node_modules/@babel/polyfill/dist/polyfill.js',
       './node_modules/phantomjs-polyfill/bind-polyfill.js',
       './scripts/__tests__/**/*.js'
     ],
--- a/package.json
+++ b/package.json
@@ -42,10 +42,10 @@
     "phantomjs-polyfill": "0.0.1",
     "phantomjs-prebuilt": "^2.1.7",
     "prop-types": "^15.6.0",
-    "react": "^0.14.3 || ^15.0.1 || ^16.0.0",
+    "react": "^15.5.4 || ^16.0.0",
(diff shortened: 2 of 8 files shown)


thathenderson:
Tried a few things, but it seems like I can't update `react-chartist` to solve peer dependency errors in Node 4 and below due to some issues between the `webpack:docs` task and `react-chartist` 0.13.1.

Not sure how to proceed with that, but at least it seems to work fine in Node 5+

dahlbyk:
FWIW, @ryanlanciaux @joellanciaux the Chartist demo on https://griddlegriddle.github.io/v0-docs/customization.html is broken:

> TypeError: root is undefined (ChartistGraph.js:177:108)
> Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `OtherPager`. See https://fb.me/react-warning-keys for more information. (react.js:19254:9)
> TypeError: Chartist[type] is not a constructor (ChartistGraph.js:129:26)

ryanlanciaux:
Ahh - wonder if we should take that out.

dahlbyk:
> Ahh - wonder if we should take that out.

Nobody seems to have noticed, so it doesn't seem worth bending over backward to manage the extra dependency just for docs. 🤷‍♂️

PR #813, opened 2018-04-29
Fix: Cannot read property 'onSort' of undefined

### LINKED ISSUE #812: Cannot read property 'onSort' of undefined
## Griddle version
1.13.0
## Expected Behavior
It renders without errors
## Actual Behavior
`Cannot read property 'onSort' of undefined` error
## Steps to reproduce
Just try to use Griddle without the local plugin.  
https://codesandbox.io/s/zk67w0x2p
## Pull request with failing test or storybook story with issue

While this step is not necessary, a failing test(s) and/or a [storybook story](https://github.com/storybooks/react-storybook) will help us resolve the issue much more easily. Please see the README for more information.

## The example works for version 1.12.0, but not for 1.13.0

dahlbyk:
## Griddle major version

1.13.0

Reconciles one-line difference between default and local `TableHeadingCellContainer`, which left the former without access to `events`.

Since those files are supposed to match as of #749, it seems reasonable to avoid double maintenance. As with [`TableHeadingCellEnhancer`](https://github.com/dahlbyk/Griddle/blob/c8130764d02ad7b85827e85c467a25c7b8a45b61/src/components/TableHeadingCellEnhancer.js), I've tried to avoid a breaking change for folks who are depending on `LocalPlugin` exporting `components.TableHeadingCellContainer` by re-exporting the default component.

## Why these changes are made

Fixes #812 (cc @bsalex)

## Are there tests?

Storybook no longer breaks. Anyone know if Storybook can be run through CI?


ryanlanciaux:
This looks good to me! Not sure about storybook through CI -- would be handy almost to be able to run those as cypress tests haha.

Issue #836, opened 2018-10-14
when <RowDefinition> children contains <ColumnDefinition> and an array of <ColumnDefinition>


JaosnHsieh:
## Griddle version
1.13.1
## Expected Behavior
Work as flat array of <ColumnDefinition> is flat elements.
## Actual Behavior
TypeError: Cannot read property 'id' of undefined
## Steps to reproduce
const data = [
  {
    name: "data1",
    name1: 111,
    name2: 222
  },
  {
    name: "data2",
    name1: 333,
    name2: 444
  }
];
class App extends Component {
  render() {
    return (
      <Griddle data={data} plugins={[plugins.LocalPlugin]}>
        <RowDefinition>
          <ColumnDefinition id={"name"} title={"name"} />
          {["name1", "name2"].map(c => {
            return <ColumnDefinition key={c} id={c} title={c} />;
          })}
        </RowDefinition>
      </Griddle>
    );
  }
}
## Pull request with failing test or storybook story with issue

While this step is not necessary, a failing test(s) and/or a [storybook story](https://github.com/storybooks/react-storybook) will help us resolve the issue much more easily. Please see the README for more information.

PR #837, opened 2018-10-14
when <RowDefinition> children contains <ColumnDefinition> and an arra…

JaosnHsieh:
…y of <ColumnDefinition>

## Griddle major version
1.13.1
Wish <RowDefinition> could have props.Children is <ColumnDefinition> and array of <ColumnDefinition> at the same time.
## Why these changes are made
#836 
## Are there tests?
No

### DIFF
--- a/src/utils/columnUtils.js
+++ b/src/utils/columnUtils.js
@@ -25,7 +25,14 @@ export function getColumnProperties(rowProperties, allColumns=[]) {
     // build one object that contains all of the column properties keyed by id
     children.reduce((previous, current, i) => {
       if (current) {
-        previous[current.props.id] = {order: offset + i, ...current.props};
+        if(Array.isArray(current)){
+          current.forEach((c)=>{
+            previous[c.props.id] = {order: offset + i, ...c.props};
+          });
+        }else{
+          previous[current.props.id] = {order: offset + i, ...current.props};
+        }
+        
       }
       return previous;
     }, columnProperties);


src/utils/columnUtils.js
```diff
@@ -25,7 +25,14 @@ export function getColumnProperties(rowProperties, allColumns=[]) {
     // build one object that contains all of the column properties keyed by id
     children.reduce((previous, current, i) => {
       if (current) {
-        previous[current.props.id] = {order: offset + i, ...current.props};
+        if(Array.isArray(current)){
+          current.forEach((c)=>{
+            previous[c.props.id] = {order: offset + i, ...c.props};
```
dahlbyk: This will end up producing multiple columns that share an `order` value, which can produce unexpected behavior.

src/utils/columnUtils.js
```diff
@@ -25,7 +25,14 @@ export function getColumnProperties(rowProperties, allColumns=[]) {
     // build one object that contains all of the column properties keyed by id
     children.reduce((previous, current, i) => {
```
dahlbyk: Would a simpler solution be to `_.flatten(children).reduce(...)` here?

src/utils/columnUtils.js
JaosnHsieh: change to _.flatten also solve this problem, thx.

src/utils/columnUtils.js
JaosnHsieh: it's better!

JaosnHsieh:
just updated to current version and add sotrybook example

PR #839, opened 2018-11-05
Upgrade recompose to ^0.25.1

mDibyo:
## Griddle major version
v1

Upgrade recompose to ^0.25.1

## Why these changes are made
This is the first version of recompose "officially" compatible with React 16.
(https://github.com/acdlite/recompose/pull/530#issuecomment-332285889)

Griddle react seems to work fine in React 16 without this change.

#823 would achieve the same result, but seems to have failing CI tests.
This is minimal change required to resolve warning `"griddle-react > recompose@0.21.2" has incorrect peer dependency` from npm/yarn.

## Are there tests?
No additional tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "max-safe-integer": "^1.0.0",
     "prop-types": "^15.5.8",
     "react-redux": "^5.0.6",
-    "recompose": "^0.21.2",
+    "recompose": "^0.25.1",
     "redux": "^3.5.2",
     "reselect": "^2.5.3"
   },


mDibyo:
CI is failing in `npm run check-ts` step, due to unrelated reasons.
#840 fixes CI.

mDibyo:
After merging in #840, specs pass

ibrahima:
Hi! Is there any chance that this could be merged? It looks like specs pass now and it seems like there should not be any significant changes: https://github.com/acdlite/recompose/releases?after=v0.25.1

I think technically Griddle can be used on React 16 already, but you will get warnings during npm/yarn install about incorrect peer dependency versions.

Thanks!

PR #866, opened 2021-12-08
Fix: setFilter reaction to componentWillReceiveProps in Griddle

dearsaturn:
Ignore this!


dearsaturn:
Whoops, this is the wrong repository. This is for an internal fork.

